PR #1, opened 2015-06-02
Monkey patch slack-client sendMessage method

Kami:
This is a temporary work-around until our pull request gets merged.

The difference between our and upstream `sendMessage` method is that we also send `parse` attribute letting slack API know to not parse and format the message.


manasdk:
+1

enykeev:
PR for the reference: slackhq/node-slack-client#42

PR #3, opened 2015-06-02
Hubot test with a mock adapter

manasdk:
* Added a mock adapter which is used to bootstrap the tests
* Trivial tests to validate setup functionality

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,13 +29,17 @@
     "hubot": "2.x"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


manasdk:
I am going to stop this line of testing here and moving to an e2e testing approach.

manasdk:
Damn it stupid editor went ahead and changed everything automatically. Too much pain to reverse those changes right now. WIll try to avoid formatting changes in the future.

PR #5, opened 2015-06-04
Include link to the WebUI history view if ST2_WEBUI_URL environment variable is set

Kami:
This pull request adds support for including a link to the execution details in WebUI if the ``ST2_WEBUI_URL`` environment variable is configured.

screenshot from 2015-06-04 17 41 03

screenshot from 2015-06-04 17 41 23


scripts/stackstorm.js
```diff
@@ -162,6 +169,37 @@ module.exports = function(robot) {
     );
   };
 
+  var getExecutionHistoryUrl = function(execution_id) {
```
Kami: I know this is an anti-pattern since it makes things hard to test and less re-usable, but I just followed the existing style.

After this PR is merged, I will open a separate one where I will refactor stackstorm.js module. I will move most of the functions to the top-level (outside of the closure), export them and add tests. This will also make code more re-usable.

Kami: See #6

jfryman:
In a WF, will each task also drop the URL?

Kami:
I think so, but I would need to verify it.

PR #8, opened 2015-06-11
Fix ST2_CHANNEL description.

manasdk:
* Fix description and clarify meaning of ST2_CHANNEL

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,10 @@ specified when running hubot:
 
 * `ST2_API` - URL to the StackStorm API endpoint.
 * `ST2_CHANNEL` - Notification channel where all the notification messages
-  should be sent to. If you use Slack adapter, that's the name of the Slack
-  channel.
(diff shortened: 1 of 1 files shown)


armab:
Since `ST2_CHANNEL=hubot` by default, can we say that it's optional in README?


In [running the bot command](https://github.com/StackStorm/hubot-stackstorm#running-the-bot) there is `ST2_CHANNEL=mychannel`

```sh
ST2_AUTH_USERNAME=testu \
ST2_AUTH_PASSWORD=testp \
HUBOT_SLACK_TOKEN=token \
# not needed to set up in most of the cases
ST2_CHANNEL=mychannel \
PORT=8181 \
bin/hubot --name "st2-bot" -a slack --alias !
```

manasdk:
> Since ST2_CHANNEL=hubot by default, can we say that it's optional in README?

yes. Will add that clarification. I should really not be making any code changes this early in the morning :)

PR #9, opened 2015-06-12
Reload available commands on SIGUSR2 signal

Kami:
If user doesn't want to wait on the periodic commands reload they can now also send `SIGUSR2` signal to the bot process which will cause it to reload the commands without a restart.


manasdk:
Ok. We need some coding conventions.

https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml

I don't care what the conventions are as long as they can be automated and violations identified by a linter.

manasdk:
Merged master to remove conflicts.

PR #14, opened 2015-07-11
Break out auth in its owns module

manasdk:
* With this change the authentication code is moved into a separate module. The modularization helps testability and also keep stackstorm.js sane.
* Added test cases for authentication and while working on tests cleaned up some previous tests also and/or fixed broken tests.
* Authentication is now a little resilient at bootstrap i.e. wait for the service to initialize before giving up.
* Failure to authenticate will no longer cause hubot to exit. An error is logged and hubot continues to work - this is to support deployments on existing hubot installs which seems like a common deployment pattern.


manasdk:
Testing done -
* Validated with authentication enabled
* Validated without authentication enabled
* Validated that hubot-stackstorm does not give on the first authentication failure

manasdk:
Merging. No peer-reviewer around to +1 this change so assuming that my local testing is sufficient.

PR #18, opened 2015-07-15
Refactor formatter

manasdk:
### Changes

Large PR with many different changes. Which got inter-mingled with each other and I lost patience to detangle the mess.

* Moved all adapter specific code into specific modules
* Dropped all tableformat and jsonformat code since it is not used anyway
* Move recipient formatting to format module as well so that adapter specific is hidden away
* Truncate message text at 500 characters. (value is configurable via environment variables)
* Hardcode the right hipchat domain.
* Slack replace \u0022 with \u201c when it notices a left quote mark. This breaks
  the command parsing on the back end. So something like run local "uanme -a" would
  not work. Normalizing brings back the functionality in slack. Ofcourse the
  downside to the approach is if a user legitemately would like to use left double
  quote mark then that would blow-up.
* Special formatting directive for hipchat.
* To have control over what get truncate and what does not send out response
  to slack as multiple messages. This provide some flexibility in message
  formatting.

### Notes on testing
Using the `slack` and `hipchat` adapters tested the following commands -
* `!run local date` - Basic test
* `!run local "uname -a"` - Issue with quotes
* `!run local "cat /tmp/Makefile"` - To validate truncation

### DIFF
--- a/lib/format_data.js
+++ b/lib/format_data.js
@@ -18,58 +18,78 @@ limitations under the License.
 "use strict";
 
 var _ = require('lodash'),
-    util = require('util'),
-    Table = require('cli-table');
+  util = require('util'),
+  utils = require('./utils.js');
 
-module.exports = function(data, formatType, logger) {
-  var tableFormatData = function(data, logger) {
-    if (!data) {
-      return "";
-    }
+var env = process.env;
 
-    data = JSON.parse(data);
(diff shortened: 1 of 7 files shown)


manasdk:
Should probably rename formatted to adapter or something. Since it is basically a service specific handler.

PR #28, opened 2015-07-29
Document ST2_MAX_MESSAGE_LENGTH

armab:
Was surprised to see that messages are truncated now.

BTW, are you sure they should be truncated to `500` by default?
Personally I would always set it to `0` for no limit. Feel like it should be default behavior with optional possibility to cut those messages.
Especially because there wasn't such limit before.

ex


jfryman:
@armab we made it default so that the formatting wouldn't be horrible with Hubot and some chat clients. We'll adjust this once we have better chat formatting capabilities.

armab:
Got it, thanks!

PR #29, opened 2015-07-29
regex not matching properly

amaline:
The regex was maching on any character "." which included spaces.  changed to be "not space" ^\s


manasdk:
@amaline Can you add a test for this in https://github.com/StackStorm/hubot-stackstorm/blob/master/tests/test-command-factory.js

jfryman:
Superseded by #30

PR #30, opened 2015-07-30
fixed regex on command factory w/tests

amaline:
cancel the previous pull request.  This one updates the test fixtures with a new test for the case when two formats match at the beginning.  e.g

alias4 fmt1 value1
alias4 fmt1 value2 word value2


Kami:
Thanks.

The change looks good to me, but to be on the safe side, we should also add some tests for strings with "key=value" parameters at the end.

Kami:
Also, on a related note - if you have two aliases which share the common prefix, you need to make sure the longer one is defined first.

At least that was the case before this PR :)

manasdk:
Ugh. Not sure why we never merged this sooner. Merging!

PR #32, opened 2015-07-31
Better error reporting if we fail to load the commands because of an exception

Kami:
Previously, if an exception was thrown (e.g. connection refused - API is offline, connection timeout, etc.). we wouldn't correctly log the reason for the loading failure (the error would just say null).

### DIFF
--- a/scripts/stackstorm.js
+++ b/scripts/stackstorm.js
@@ -106,11 +106,11 @@ module.exports = function(robot) {
 
     request.get()(
       function(err, resp, body) {
-        var parsed_body, success, error_msg_prefix;
+        var parsed_body, success, error_msg;
 
-        error_msg_prefix = 'Failed to retrieve commands from ' + env.ST2_API + ': ';
         if (err) {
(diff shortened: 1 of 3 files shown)


scripts/stackstorm.js
```diff
@@ -106,7 +106,13 @@ module.exports = function(robot) {
 
     request.get()(
       function(err, resp, body) {
-        var parsed_body, success;
+        var parsed_body, success, error_msg_prefix;
+
+        error_msg_prefix = 'Failed to retrieve commands from ' + env.ST2_API + ': ';
```
manasdk: Can move into if(err)  block

manasdk:
+1

scripts/stackstorm.js
Kami: It's actually used in two if blocks (also the one which checks for `success`).

In any case, yeah, using a module level constant or similar would make it nicer.

Kami: Pushed a change to use util.format.

PR #34, opened 2015-08-12
RFR: Rename notification channel to notification route.

lakshmi-kannan:
Related: https://github.com/StackStorm/st2/pull/1817


manasdk:
+1

PR #44, opened 2015-10-10
Option for the colors and putting back the username when not a whisper

thomaspicquet:
slack

Added options for the slack attachement colors and I put back the username in the message


jfryman:
Another fantastic UX enhancement. Thanks for the contribution! 🙇

PR #46, opened 2015-10-14
Refactor message format into adapter specific classes

manasdk:
* Also tests now since it can be unit tested
* Truncation of messages posted to slack is no longer necessary.

Validated changes with `slack` and `hipchat`. Looks like all the basic stuff is still working fine after my change.


manasdk:
cc thomaspicquet

thomaspicquet:
Looks good! @manasdk thanks!

PR #49, opened 2015-10-29
Switch to using st2client

enykeev:


enykeev:
Requires st2client.js 0.4.1. ETA: EoD PST

enykeev:
I agree that we need to kind of take a step back and see the whole picture. At the same time, I really see a value for us to switch to stream as it is definitely the better communication channel than calling a webhook.

The thing I need to know is whether we use ST2_ROUTE somewhere else at the moment. If we don't, it seems like an extremely easy way for us to support both channels for the time being and figure out what we want at the end by fiddling in code rather than having a scientific discussion. If we don't, we should merge that one and start playing with it. It would unblock me on getting custom formatting and it won't block us from using and improving the old mechanics in case we find a good reason to stick with that.

You're saying we want to be able to skip some notifications. Ok, lets do that. With stream we can make it by a) adding a flag in execution context or b) by emitting specific event for chatops results (same way we do now with webhook, short of webhook itself). How exactly do you want to differentiate between actions that will or won't notify?

emedvedev:
I'm +1 on merging this: ChatOps works right now, sure, but it still needs some love from us, and since we have time to focus on this right now, I'd say streams is a good place to start. Some minor things may regress for a while, but we can get this polished before the next release, and nobody will notice a thing unless they're running bleeding edge. This PR is great to get the ball rolling.

Re actions that will or won't notify: I think we should treat chatops-triggered executions and other executions as we do now, but extend it a bit:

— notify of every execution triggered by a chatops command UNLESS specified otherwise in an alias YAML definition (I have this working on my local instance, that's what @enykeev referenced before)

— not notify of anything else UNLESS there's a custom rule a user created

manasdk:
> ChatOps works right now, sure, but it still needs some love from us, and since we have time to focus on this right now, I'd say streams is a good place to start. Some minor things may regress for a while, but we can get this polished before the next release, and nobody will notice a thing unless they're running bleeding edge. This PR is great to get the ball rolling.

I agree and admire the spirit but I do not think we have discussed the use-cases. It is not like this approach was not available when we/I originally designed it the way it is right now. Like I said lets have a clear discussion on what and where and then move this PR forward based on a shared understanding.

emedvedev:
That's nothing if not sensible, I agree.

PR #55, opened 2015-11-11
Fix custom host

enykeev:


Kami:
Good catch, would also be good to add some tests (I thought we had some already for this, there was a parseUrl or smth function, but maybe we don't) :)

PR #59, opened 2015-11-16
Dumb down part 2: ack messages

emedvedev:
Another part of “dumbing down”: instead of controlling ack message format on the hubot-stackstorm side, we move it to backend (because of the new `format` requirements), and use hubot to output whatever it got from API.

Must not be merged before the respective API changes are made, otherwise it will always output execution id as ack.

### DIFF
--- a/scripts/stackstorm.js
+++ b/scripts/stackstorm.js
@@ -195,7 +195,9 @@ module.exports = function(robot) {
         // Until aliasexecution endpoint didn't get patched with proper status and output, work
         // around this curious design decision.
         if (err.status === 200) {
-          msg.send(err.message);
+          if (err.message && err.message.length) {
+            msg.send(err.message);
+          }
         } else {
           msg.send(util.format('Error: %s', err.message));
         }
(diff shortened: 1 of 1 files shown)


emedvedev:
Sadly, I had to get rid of our beloved randomly picked fun messages to accomplish this. It's now up to @enykeev to recreate this in the backend. :stuck_out_tongue:

manasdk:
Makes no sense to me. Why does this goto the back-end?

The ACK must be fully managed in hubot. If we are not doing so we are doing something wrong.

emedvedev:
We want to manage ack formatting the same way we manage result formatting: having support for Jinja templating, for one, and inserting executions params. There's no point in managing one on the front-end and the other one on the back-end, it's unnecessary duplicating.

In our current state, we have to ask back-end for execution id, so there's still no ack without talking to a server. If this is the case, why is pushing formatting to back-end such a bad idea?

emedvedev:
Again, ACK is _not_ fully managed in Hubot right now. It requires a server request.

enykeev:
I agree that custom ACK responses should be implemented on backend simply because of jinja. On the other hand, I think we should leave random messages be and live them on a client for situations when there's no custom ack. We don't need them to be Jinja and it's up to bot developer to decide whether he wants it to be fun or not.

enykeev:
It would be much more helpful to me if I could get this commit as a list of suggestions...

emedvedev:
How will the bot tell if what he got is a custom ack or not? In any case, there's no point in arguing before the backend impl is there.

emedvedev:
Trumped by #64

PR #60, opened 2015-11-16
Pretext delimiter

emedvedev:
As discussed, I’m introducing a pretext delimiter for separating preformatted text (served as attachment / code / backticks depending on adapter) from plaintext in the response. There are multiple use-cases for this, and Slack in particular dislikes attachment-only responses, excluding them from notifications, etc.

Default adapter falls back to concat, separating with a newline.

I went with `::` as a delimiter because @lakshmi-kannan was the first with a suggestion. Personally I’d prefer `~~~`, and @manasdk suggested `::.`. I honestly don’t care, so if either of you have a preference based on a convention, previous experience or a very strong gut feeling, speak up, I’ll change it. :)

As for using the delimiter at all: it’s hacky, we all agreed on that, but required very little modification and on `hubot-stackstorm` part only. Once we get to refactoring `format_data` and `post_data` (soon, I hope) we can think about replacing this with something bulletproof.


emedvedev:
Tests will fail, because it partly depends on the "dumbing down" PRs. I'm going to merge them into one branch before merging into master. Separate PRs are solely for the purpose of having separate discussions.

lib/post_data.js
```diff
@@ -30,39 +30,36 @@ function SlackDataPostHandler(robot, formatter) {
...
   attachment_color = env.ST2_SLACK_SUCCESS_COLOR;
   if (data.message.indexOf("status : failed") > -1) {
     attachment_color = env.ST2_SLACK_FAIL_COLOR;
   }
-  var text = "";
-  // PM user, notify user, or tell channel
-  if (!data.whisper) {
-    text = util.format('%s :', data.user);
+
+  splitted = utils.splitMessage(this.formatter.formatData(data.message));
```
manasdk: Use `split` i.e. `splitMessage` would be nicer.

manasdk:
Don't care about the specific delimiter. I actually am not sure I like the delimiter approach much anyway. We can revisit later.

PR #61, opened 2015-11-18
Improvements in auth and error reporting

enykeev:


scripts/stackstorm.js
```diff
@@ -76,15 +76,19 @@ env.ST2_MAX_MESSAGE_LENGTH = parseInt(env.ST2_MAX_MESSAGE_LENGTH || 500, 10);
...
-    "I'll take it from here! Your execution ID for reference is %s",
-    "Got it! Remember %s as your execution ID",
-    "I'm on it! Your execution ID is %s",
-    "Let me get right on that. Remember %s as your execution ID",
-    "Always something with you. :) I'll take care of that. Your ID is %s",
-    "I have it covered. Your execution ID is %s",
-    "Let me start up the machine! Your execution ID is %s",
-    "I'll throw that task in the oven and get cookin'! Your execution ID is %s",
-    "Want me to take that off your hand? You got it! Don't forget your execution ID: %s"
+  "I'll take it from here! Your execution ID for reference is %s",
```
manasdk: Didn't we remove these?

enykeev: Nope, I've reverted it. Too soon =)

jfryman: :heart:

scripts/stackstorm.js
```diff
@@ -76,15 +76,19 @@ env.ST2_MAX_MESSAGE_LENGTH = parseInt(env.ST2_MAX_MESSAGE_LENGTH || 500, 10);
...
+  "I'm on it! Your execution ID is %s",
+  "Let me get right on that. Remember %s as your execution ID",
+  "Always something with you. :) I'll take care of that. Your ID is %s",
+  "I have it covered. Your execution ID is %s",
+  "Let me start up the machine! Your execution ID is %s",
+  "I'll throw that task in the oven and get cookin'! Your execution ID is %s",
+  "Want me to take that off your hand? You got it! Don't forget your execution ID: %s"
+];
+
+var ERROR_MESSAGES = [
```
jfryman: YES!!!!!

enykeev: We will keep it anyway. We may want to move it to st2 side, though I think the decision whether bot should be fun or not should be on one implementing the bot.

emedvedev: Ideally it should belong here, but back-end has to give a clear indication when the ack text is pre-formatted and the user wants it to remain as is.

PR #64, opened 2015-11-20
New ack formatting

emedvedev:
Supports the new object format passed by AliasExecution API (https://github.com/StackStorm/st2/pull/2229).

1. Correctly gets the execution id from context.
2. Uses actionalias from the API.
3. If `message` (jinja-formatted string) is specified, uses it.
4. If there’s no `message`, reverts to the standard format (with fun and all that).
5. Extra Firefly reference for bonus points from @jfryman.


emedvedev:
Should only be merged after the corresponding API PR is reviewed, tested and merged, otherwise will not work.

PR #66, opened 2015-11-26
Ignore empty messages

emedvedev:
Don’t post message if there’s no message body.

### DIFF
--- a/scripts/stackstorm.js
+++ b/scripts/stackstorm.js
@@ -271,7 +271,9 @@ module.exports = function(robot) {
         data = req.body;
       }
 
-      postDataHandler.postData(data);
+      if (data && data.message) {
+        postDataHandler.postData(data);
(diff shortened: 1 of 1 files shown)


scripts/stackstorm.js
```diff
@@ -302,7 +304,9 @@ module.exports = function(robot) {
           data = e.data;
         }
 
-        postDataHandler.postData(data);
+        if (data && data.message) {
```
manasdk: Is there any other envelope information that you can post in this case? If not a sense of nothing happened is left with the user.

emedvedev: > nothing happened

That's exactly the point: if an empty string is passed, then nothing happened. Until we can use Mistral to avoid `post_message` execution in the workflow or rework announcement runner, that's the most we can do.

PR #74, opened 2015-12-02
Prevent disabled aliases from showing up in help

enykeev:


emedvedev:
+1 for this release, but ideally this filtering should happen in the backend, don't you think?

enykeev:
No, frankly, I don't think so. As I said earlier, we don't filter disabled actions or rules on backend, why should we do that for aliases. It's not like we will have a thousands of them disabled and it will became a question of traffic. Sorting them client side would allow to show them with some additional flag for the purpose of discovering and enabling them right from chat without switching context.

emedvedev:
Because when someone disabled an alias, they explicitly don't want it to work—and potentially to be exposed at all, even to a bot. I don't insist though, see your point.

enykeev:
> they explicitly don't want it to work

that's an opinion

emedvedev:
That's why I don't insist. :) Wanted to start a discussion and get another opinion rather than actually change something.

manasdk:
The filter should not be implicit. Infact  hubot-stackstorm can query only for those aliases that are enabled. If that filter is not supported then this is a valid choice as it is client responsibility to ignore a disabled alias. In some contexts there is value in viewing disabled aliases e.g. CLI.

`/aliases?enabled=True`.

enykeev:
Sorry @manasdk, not working for me. We'll refactor it next time.

```
$ http https://192.168.10.11:443/api/v1/actionalias\?enabled\=true\&x-auth-token\=19badad7716743098a03496df63e2cea --verify no
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type,Authorization,X-Auth-Token,St2-Api-Key,X-Request-ID
Access-Control-Allow-Methods: GET,POST,PUT,DELETE,OPTIONS
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: Content-Type,X-Limit,X-Total-Count,X-Request-ID
Connection: keep-alive
Content-Length: 4242
Content-Type: application/json; charset=UTF-8
Date: Fri, 04 Dec 2015 08:27:05 GMT
Front-End-Https: on
Server: nginx/1.4.6 (Ubuntu)
X-Content-Type-Options: nosniff
X-Request-ID: 2189a0c5-700c-4494-84fd-3038ab05ea16
X-Total-Count: 12

[
    {
        "ack": {
            "format": "requesting..."
        },
        "action_ref": "core.http",
        "description": "Make http GET request",
        "enabled": false,
        "formats": [
            "http get {{ url }}"
        ],
        "id": "566131eb1c3d9a189f3b2d31",
        "name": "get",
        "pack": "chatops",
        "ref": "chatops.get",
        "result": {
            "format": "done, here it is: {{ execution.result.body }}"
        },
        "uid": "action:chatops:get"
    },
    {
        "ack": {},
        "action_ref": "hubot.branch",
        "description": "Show the currently deployed hubot branch",
        "enabled": true,
        "formats": [
            "hubot branch",
            "hubot branch on {{hosts}}"
        ],
        "id": "566132681c3d9a1e92409472",
        "name": "hubot_branch",
        "pack": "hubot",
        "ref": "hubot.hubot_branch",
...
```

manasdk:
> Sorry @manasdk, not working for me. We'll refactor it next time.

Thanks for trying. I will fix the backend APIs.

PR #76, opened 2015-12-02
Tweak help generation

enykeev:
Due to how hubot-help is implemented, it automatically replaces every mention of 'hubot' with bot alias.

I personally think this was a bad idea on hubot's part and it seems especially awful in our case, but we've decided to follow this convention for the time being instead of re-implementing help module.

### DIFF
--- a/lib/command_factory.js
+++ b/lib/command_factory.js
@@ -70,7 +70,7 @@ CommandFactory.prototype.addCommand = function(command, name, format, action_ali
     command: command
   };
 
-  compiled_template = _.template('${robotName} ${command}');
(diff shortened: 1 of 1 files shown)


jfryman:
@enykeev Think that we could inject a "helpful message" that says something along the lines of "replace hubot with your bots name"

enykeev:
Not really sure where we should inject this message. All things consider, I would rather prefer to rewrite `hubot-help` and let the world burn, but I guess we can always do that later.

jfryman:
@emedvedev :fire: :earth_americas:. WFM... don't burn the energy if it's not helpful.

PR #87, opened 2016-02-11
Self-check

emedvedev:
A basic self-check script that would connect to the container and to StackStorm. Right now it checks for the following:

1. Hubot is installed and the container is running
2. Hubot-stackstorm is installed inside the container
3. There are enabled StackStorm aliases in the system
4. `chatops.notify` rule is present
5. `chatops.notify` rule is enabled

I’ll also make it peek inside Hubot to check if it can actually get info from st2, and maybe do a `post_message`, too.

Also, my bash-fu is really bad, so feel free to fix or prettify things. :P

cc @lakshmi-kannan


emedvedev:
PAIN PAIN A LOT OF PAIN. That's what went into trying to get Hubot to reply in the interactive shell from outside of its Docker container today. We'll overcome!

lakshmi-kannan:
>Also, my bash-fu is really bad, so feel free to fix or prettify things. :P

heh. I don't know man. Looks good so far!

emedvedev:
This is pretty neat!

```

Starting the Nine-Step Hubot Self-Check Program
===============================================

Step 1: Hubot is running.
Step 2: Hubot-stackstorm is installed (hubot-stackstorm@0.3.4).
Step 3: StackStorm has aliases that are registered and enabled.
Step 4: Chatops.notify rule is present.
Step 5: Chatops.notify rule is enabled.
Step 6: Hubot responds to the help command.
Step 7: Hubot loads commands from StackStorm.
Step 8: chatops.post_message execution succeeded.
Step 9: chatops.post_message has been received.

===============================================

Everything seems to be fine!

Hubot is working, StackStorm commands are loaded normally
and messages from StackStorm are getting through.

If you can't see the bot in your chat at this point,
the most probable cause is incorrect login credentials.

Check that your bot is using the right credentials to log in.
If you installed StackStorm with the All-In-One Installer,
the Hubot init script is located at:

/etc/init.d/docker-hubot

If you're still having trouble, gist the log files
and come see us in our Slack community:
https://stackstorm.com/community-signup

You can access Hubot logs with:
docker logs hubot

StackStorm logs are stored in:
/var/log/st2/

```

Please someone else test the script, and then we can merge and use this.

emedvedev:
This script makes a few assumptions:

1. `st2` points to the right StackStorm instance and the user has already authenticated.
2. The user can run `sudo` (it's required for docker).
3. Hubot is installed inside a Docker container. It's the case with every post-1.2 install, and if someone is using their own Hubot, it's up to them to debug it. I don't think we can (or should) support every case.

Nothing wrong with any of this, just FYI.

PR #88, opened 2016-02-17
Allow specifying color for Slack messages

agaffney:
This commit allows specifying a color for the Slack "attachment". It falls back on the old logic if no color is specified with the message.


agaffney:
This is meant to go with https://github.com/StackStorm/st2/pull/2510

manasdk:
LGTM

Will wait for @emedvedev to take a look.

emedvedev:
LGTM. Merging.

PR #97, opened 2016-03-07
Hotfix: avoid duplicating help entries

emedvedev:
In some cases with `display/representation` aliases help entries might
duplicate. This is a simple fix checking if the entry is already in the
array.

### DIFF
--- a/lib/command_factory.js
+++ b/lib/command_factory.js
@@ -73,15 +73,15 @@ CommandFactory.prototype.addCommand = function(command, name, format, action_ali
   compiled_template = _.template('hubot ${command}');
   command_string = compiled_template(context);
 
-  if (!flag || flag === utils.DISPLAY) {
-    this.robot.commands.push(command_string);
-  }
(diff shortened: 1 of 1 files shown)


pixelrebel:
This didn't fix my problem.  My issue arrises when st2 sends commands to hubot.  
When hubot first starts, it receives the set of commands from st2 and caches the help dialogues.  Every two minutes hubot receives the updated list of commands from st2.  Instead of _updating_ the help list, it _appends_ to the list.  So after two minutes there are two of each command after four minutes there are three of each command and so on.  This only happens for action aliases that use the `display/representation` method of formatting.

emedvedev:
That's exactly what this PR is meant to address: it checks if the new command is already in the list before appending to it on every update. Can you check the update applied correctly? It fixed the issue on my instance.

pixelrebel:
Okay, looks like I didn't update properly.  I updated my entire hubot installation and that seems to do the trick.

emedvedev:
Awesome. Thanks for catching the issue!

PR #99, opened 2016-03-12
Match commands ignoring case

emedvedev:
As planned originally, commands are now matched in a case-insensitive
way; it’s achieved with the `i` flag instead of converting to
lowercase, because the latter is error-prone.

Fixes https://github.com/StackStorm/st2/issues/2566


Kami:
LGTM, but let's please document this in the alias / ChatOps section of the documentation so there will be no surprises.

PR #102, opened 2016-03-15
[RFR] Native support for two-factor auth

emedvedev:
Goes together with https://github.com/StackStorm/st2contrib/pull/444.

This PR adds native support for two-factor auth through the Duo pack. Support can be enabled by setting `HUBOT_2FA` environment variable and having `extra:security:twofactor` flag in an alias. 

It works through “freezing” an execution attempt until a `2fa` announcement for a successful auth with the correspondent ID is caught in the stream.

The change is fully backwards-compatible; later I’ll add the env variable (commented out) to `st2chatops.env`, thoroughly documenting the case as well.

### DIFF
--- a/lib/format_data.js
+++ b/lib/format_data.js
@@ -73,7 +73,7 @@ HipChatFormatter.prototype.formatData = function(data) {
 
 HipChatFormatter.prototype.formatRecepient = function(recepient) {
   var robot_name = env.HUBOT_HIPCHAT_JID.split("_")[0];
-  var hipchat_domain = (env.HUBOT_HIPCHAT_XMPP_DOMAIN == 'btf.hipchat.com') ?
+  var hipchat_domain = (env.HUBOT_HIPCHAT_XMPP_DOMAIN === 'btf.hipchat.com') ?
                        'conf.btf.hipchat.com' : 'conf.hipchat.com';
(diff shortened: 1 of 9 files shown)


emedvedev:
A thought I'm still pondering is skipping the manual `! 2fa UUID` confirmation and just sending the 2fa request straight away. It's faster and easier for the user, less copy-pasting and typing, but could be a bit unexpected. :)

jjm:
I really like this approach, makes it easier to enable for single actions and will remove bloat caused by adding extra workflows to just call `duo.auth_auth`

Kami:
We didn't have much discussion around this and this approach is a bit duo specific, but I'm personally OK with this approach in the short term.

As far as the code goes - this is an important security feature so we definitely need good (unit) test cases for it.

emedvedev:
Definitely. I'll add tests as soon as I figure out if I can safely remove the extra step.

emedvedev:
> this approach is a bit duo specific

I re-worked it a bit, so now it's not directly tied to Duo. You can specify any action to trigger 2FA, from any pack; it just needs to have the same parameter set and send the same announcement on success.

emedvedev:
Added tests to check if FA should be enabled/disabled in general as well as for specific aliases. As the feature grows, we'll need more tests for sure.

PR #104, opened 2016-03-23
bypass room check for spark bot

tonybaloney:


scripts/stackstorm.js
```diff
@@ -267,6 +267,9 @@ module.exports = function(robot) {
       room = String(msg.message.user.thread_id);
       name = msg.message.user.name[0];
     }
+    if (robot.adapterName === "spark") {
+      room = "spark" // unsupported
```
manasdk: Can you add a comment to explain this special value.

scripts/stackstorm.js
```diff
@@ -267,6 +267,10 @@ module.exports = function(robot) {
       room = String(msg.message.user.thread_id);
       name = msg.message.user.name[0];
     }
+    if (robot.adapterName === "spark") {
+      room = String(msg.message.user.room);
```
manasdk: why does this need to cast to a String while msg.message.user.name does not

tonybaloney: ok @manasdk updated.

scripts/stackstorm.js
```diff
@@ -267,6 +267,10 @@ module.exports = function(robot) {
       room = String(msg.message.user.thread_id);
       name = msg.message.user.name[0];
     }
+    if (robot.adapterName === "spark") {
+      room = msg.message.user.room;
+      name = msg.message.user.name
```
manasdk: also missing ;

PR #106, opened 2016-04-13
Prettify error messages on incorrect executions

emedvedev:
Error messages that Hubot posts on incorrect executions (the most common case is schema validation error in an alias) were previously sent to chat as plaintext, which is messy since the messages tend to get quite long. 

This PR makes those messages go through the same formatting process as our usual result messages do: e.g. in Slack it would result in prettified and truncated attachment with a red-colored strip.

Thanks to @Kami for ranting about this convincingly enough. 😛 

screenshot 2016-04-13 07 45 13

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -88,11 +88,38 @@ function enable2FA(action_alias) {
          action_alias.extra.security.twofactor !== undefined;
 }
 
+function normalizeAddressee(msg) {
+  var name = msg.message.user.name;
+  if (robot.adapterName === "hipchat") {
+    // Hipchat users aren't pinged by name, they're
+    // pinged by mention_name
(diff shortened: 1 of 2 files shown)


Kami:
Nice, much better :+1:

lib/utils.js
```diff
@@ -88,11 +88,38 @@ function enable2FA(action_alias) {
          action_alias.extra.security.twofactor !== undefined;
 }
 
+function normalizeAddressee(msg, adapter) {
```
Kami: I really wish we had a better code organization - ideally have a module per adapter, etc.

Anyway, that's fine for now but something we can hopefully improve down the road :)

Kami:
LGTM, +1

PR #111, opened 2016-04-21
Separate long Slack attachment into chunks

### LINKED ISSUE #108: Long messages to Slack are truncated
If `chatops.post_message` is given a sufficiently long message, it is truncated, and its length does not seem to be adjustable. I tested via the [Slack documentation-based tester](https://api.slack.com/docs/formatting/builder), and the full message is retained. It appears to be truncated at some point by hubot-stackstorm.

Setting `ST2_MAX_MESSAGE_LENGTH` had no effect. I believe that is the case because of [this comment](https://github.com/StackStorm/hubot-stackstorm/blob/f2dfdd82f68a27e73f35fff5d542a202b0d0d08b/lib/format_data.js#L38).

emedvedev:
Fixes #108.

As reported by @jmound, long attachments (over 8000 characters long to be precise) are cut by Slack without an error or exception, so we have to catch those and split into multiple separate attachments. I couldn’t find a reference to this behavior in Slack docs, but could find workarounds in other’s Slack-involving code, so it’s seems to be a common method of dealing with the problem.

Long plaintext messages are split by Slack automatically.

Extra formatting (colors, etc.) is added to every chunk. 

Plaintext (where present) is only added to the first chunk.

### DIFF
--- a/lib/post_data.js
+++ b/lib/post_data.js
@@ -71,7 +71,7 @@ SlackDataPostHandler.prototype.postData = function(data) {
       };
       robot.emit('slack-attachment', attachment);
       if (chunks.length > ++i) {
-        sendChunk(i);
+        setTimeout(function(){ sendChunk(i); }, 300);
       }
     };
     sendChunk(0);
--- a/lib/post_data.js
+++ b/lib/post_data.js
@@ -52,20 +52,24 @@ SlackDataPostHandler.prototype.postData = function(data) {
   split_message = utils.splitMessage(this.formatter.formatData(data.message));
 
   if (split_message.text) {
-    attachment = {
+    var content = {
       color: attachment_color,
-      text: split_message.text,
       "mrkdwn_in": ["text", "pretext"],
-      fallback: split_message.text
     };
     if (data.extra && data.extra.slack) {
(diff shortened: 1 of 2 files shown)


lib/post_data.js
```diff
@@ -52,20 +52,29 @@ SlackDataPostHandler.prototype.postData = function(data) {
...
+      for (var attrname in data.extra.slack) { content[attrname] = data.extra.slack[attrname]; }
     }
-    this.robot.emit('slack-attachment', {
-      channel: recipient,
-      text: pretext + split_message.pretext,
-      content: attachment
-    });
+    var robot = this.robot;
+    var chunks = split_message.text.match(/[\s\S]{1,7900}/g);
+    var sendChunk = function (i) {
```
Kami: I think it would be good to make this a top-level function so it's easier / possible to test it.

To make testing easier I would probably also get rid of some global args and pass them to (and return from) the new function instead.

emedvedev: Sinon is actually great with test cases such as this one (even does the neat fake timer thing). I updated the tests to check that the long messages are split and `robot.emit` is called twice.

jmound:
👍

PR #112, opened 2016-04-24
Allow multiple extra attachments

dbarrosop:
This PR does some modifications to how attachments work on slack so you can send multiple attachments. You can do things like:

```
result:
  format: "{{ execution.result.stdout }} {~}"
  extra:
    slack:
      attachments:
        - color: "#1100ff"
          fallback: "Required plain-text summary of the attachment number 2."
          pretext: "Optional text that appears above the attachment block number 2"
          author_name: "Bobby Tables number 2"
          author_link: "http://flickr.com/bobby/"
          author_icon: "http://flickr.com/icons/bobby.jpg"
          title: "Slack API Documentation number 2"
          title_link: "https://api.slack.com/"
          text: "{{ execution.result.stdout }}"
          fields:
            - title: "Priority"
              value: "Low"
              short: false
          image_url: "http://my-website.com/path/to/image.jpg"
          thumb_url: "http://example.com/path/to/thumb.png"
        - color: "#ffff00"
          fallback: "Required plain-text summary of the attachment number 3."
          pretext: "Optional text that appears above the attachment block number 3"
          author_name: "Bobby Tables number 3"
          author_link: "http://flickr.com/bobby/"
          author_icon: "http://flickr.com/icons/bobby.jpg"
          title: "Slack API Documentation number 3"
          title_link: "https://api.slack.com/"
          text: "Optional text that appears within the attachment number 3"
          fields:
            - title: "Priority"
              value: "Low"
              short: false
          image_url: "http://my-website.com/path/to/image.jpg"
          thumb_url: "http://example.com/path/to/thumb.png"
```

Which will actually write ```{{ execution.result.stdout }}``` and the attachments after it or you can do:

```
result:
  extra:
    slack:
      content:
          color: "#1100ff"
          fallback: "Required plain-text summary of the attachment number 2."
          pretext: "Optional text that appears above the attachment block number 2"
          author_name: "Bobby Tables number 2"
          author_link: "http://flickr.com/bobby/"
          author_icon: "http://flickr.com/icons/bobby.jpg"
          title: "Slack API Documentation number 2"
          title_link: "https://api.slack.com/"
          text: "{{ execution.result.stdout }}"
          fields:
            - title: "Priority"
              value: "Low"
              short: false
          image_url: "http://my-website.com/path/to/image.jpg"
          thumb_url: "http://example.com/path/to/thumb.png"
```

Which does exactly the same that as the current implementation of attachment. Finally, you can also combine `content` and `attachments`.

`content` is just supported to emulate the current implementation but I think it is not necessary, `attachments` should be probably good enough.

What do you think?

### DIFF
--- a/lib/post_data.js
+++ b/lib/post_data.js
@@ -50,7 +50,10 @@ SlackDataPostHandler.prototype.postData = function(data) {
   }
 
   split_message = utils.splitMessage(this.formatter.formatData(data.message));
-
+  var attachments = [];
+  if (data.extra && data.extra.slack && data.extra.slack.attachments) {
+	   for(var i=0, len=data.extra.slack.attachments.length; i < len; i++) { attachments.push(data.extra.slack.attachments[i]); }
+  }  
   if (split_message.text) {
     attachment = {
       color: attachment_color,
@@ -58,18 +61,21 @@ SlackDataPostHandler.prototype.postData = function(data) {
       "mrkdwn_in": ["text", "pretext"],
       fallback: split_message.text
     };
-    if (data.extra && data.extra.slack) {
-      for (var attrname in data.extra.slack) { attachment[attrname] = data.extra.slack[attrname]; }
+    if (data.extra && data.extra.slack && data.extra.slack.content) {
+      for (var attrname in data.extra.slack.content) { attachment[attrname] = data.extra.slack.content[attrname]; }
     }
-    this.robot.emit('slack-attachment', {
-      channel: recipient,
-      text: pretext + split_message.pretext,
-      content: attachment
-    });
+    attachments.unshift(attachment);
   } else {
     this.robot.messageRoom.call(this.robot, recipient, pretext + split_message.pretext);
   }
+    if ( attachments.length > 0 ) {
+      this.robot.emit('slack-attachment', {
+        channel: recipient,
+        content: attachments
+     });    	
+    }  
 };
+
 
 /*
   HipchatDataPostHandler.


emedvedev:
Let's start. Since it's an important feature, and one we'll have to support in the future, we need to make sure there's a consensus and all use cases are considered.

A couple notes:

1. I'm fine with dropping `extra.slack.content` (your second example) in favor of having just one element in `attachments`.

2. You lost the plaintext part of the message: https://github.com/StackStorm/hubot-stackstorm/pull/112/files#diff-28d98f2dc87688965ff5bcbbcb2e3548L66

Now, here's my short-ish design proposal:

1. As you suggested, we add an `attachments` field to `extra.slack`, which would work as you described.

2. If `attachments` is present, the plaintext part (`format`) gets printed in full, without the `{~}` separator taking effect, going before the first attachment in the list as pretext.

3. If `attachments` is not there, we get back to the old behavior, maintaining full backwards compatibility (but we'll have to cut the attachment customizing feature in favor of specifying extra parameters in `attachments`—which is not documented yet, so it's not much of a loss).

4. There should be a way to skip the attachment in the list based on a condition. I propose skipping an attachment in the list if its text is empty, but there might be cases where an attachment would only contain an image, so text being empty would be intended. Something else is a separate `condition` parameter having a jinja expression there.

5. Fallback should be filled with the attachment's text if it's not defined by the user.

6. If there's both `result.format` and an attachment pretext, they should probably concat (but I can be convinced otherwise).

--- 

Overall, this feature might be an overcomplication, but I'm fine with it since it's (1) completely optional and (2) allows for great extendability and fine-tuning. Still, I'd have to get some input on it: this should be carefully designed, and I'd really like to have someone else watch our backs here. :)

cc @manasdk @Kami @enykeev @jjm, @tonybaloney

emedvedev:
Since we're at it, something I'd really love to do, btw, is to get an attachment for every element in a list (i.e. a log tail for every failed server). Do you think it's going to be possible? If so, how do we achieve it?

emedvedev:
https://github.com/StackStorm/hubot-stackstorm/pull/111 is a conflicting PR, btw, and I plan to merge it tomorrow, but it shouldn't be hard to merge into this branch.

dbarrosop:
What's the status of this one? Do you want me to fix the conflicts and do the suggested changes? I am happy closing this PR in case you want to take over the functionality and implement it yourself. I just didn't want to open an issue and demand the feature without trying to mock a proof of concept at least ;)

Regarding your comments:

> There should be a way to skip the attachment in the list based on a condition. I propose skipping an attachment in the list if its text is empty, but there might be cases where an attachment would only contain an image, so text being empty would be intended. Something else is a separate condition parameter having a jinja expression there.

I am not sure what you mean with this. You want to have like an 'enabled' within each attachment so you can skip that element?

> If there's both result.format and an attachment pretext, they should probably concat (but I can be convinced otherwise).

My guess is that if you are using attachment you might be using the format as a preface and the attachment to show the result of something. For example, something like:

```
@some_user I finished executing this task for you. Everything went fine. Here are the results: # this line comes from the format.

attachment 1
attachment 2
...
attachment n
```

So I would rather not concat anything. Otherwise, what would be the rule? Contact format + pretext for all the attachments? Only for the first element? Feels a bit counterintuitive to me.

emedvedev:
If you have time, I'd like for you to implement the changes, at least some of them, then I can take over. I don't have enough time to do them right now, so it's up to you. :)

> I am not sure what you mean with this. You want to have like an 'enabled' within each attachment so you can skip that element?

Yes. And if a user has a Jinja template there, back-end will already take care of rendering it. So for now we should just expect `enabled` and skip it if it's defined but set to either "false" or an empty string.

> My guess is that if you are using attachment you might be using the format as a preface and the attachment to show the result of something.

That's not what I had in mind. For attachments, you can specify a `pretext` parameter: that's plaintext outside of attachment. `format` is also rendered as `pretext` to the first attachment. What if a user specifies both? I guess if we make the `format` a separate message, and not the first attachment's pretext, the problem will go away, so I suggest you go that way. :)

jjm:
Currently I can't see how we would make use of multiple attachments, as we are attempting to keep the size of information posted into ChatOps not too distributive for other users.

However it may prove useful to us in the future, so it be merged in and not making it too complex for the normal use cases is fine by me.

dbarrosop:
I am going to close this PR. I will revisit later.

Thanks for the feedback.

PR #125, opened 2016-08-03
Update README.md - Remove auth issue

lakshmi-kannan:


lakshmi-kannan:
+1 @lakshmi-kannan. You're doing good.

PR #133, opened 2017-03-06
Code Refactor

silverbp:
This refactor includes: 
- pulling all the chatops specific code into a `messaging_handler` which is then pulled in once and also each `messaging_handler` is separated out into its own file for better organization.
- changing it to support a listener per action-alias so that you can support hubot middleware plugins such as hubot-auth-middleware-ext (this was a pretty substantial change)
- adding an env variable called ST2_ALIAS_PACK_RESTRICTION which is a comma separated list that restricts it to importing only from the packs in that list (this was easier than disabling the alises I didn't want which got reenabled I believe at one point)
- added a mongodb brain persistent storage, which I took from here: https://github.com/mongodb-js/hubot-mongodb-brain and modified it to not log the document even on debug as if you have a lot of users, that goes on for a while.(the other one that comes up doesn't persist the users collection which is needed for hubot-auth, also didn't want to use redis since stackstorm doesn't install that by default and there's extra work required to make that persistent)
- if you define an extra/hubot_auth element in your action alias, it will take that data and add listener opts that the hubot-middleware-auth-ext plugin can pick up and use for authentication

if you set something like this in your env file:
```
export ST2_ALIAS_PACK_RESTRICTION=ems
export HUBOT_AUTH_ADMIN=U111111,U222222 # chatops id, in this case slack id 
export MONGODB_URL=mongodb://localhost:27017/chatops # connection string to mongo for brain storage
```

then configure your action-aliases to look like this:
```
---

name: some_action_alias
pack: some_pack
description: "this alias does some work"
action_ref: "some_pack.hello_world"
formats:
  - "hello wolrd {{ name }}"
extra:
  hubot_auth:
    roles:
      - devops
    rooms:
      - general

```

roles and/or rooms are optional although it might be pointless if you don't have either of them and still define the hubot_auth part. (it won't do auth if you leave it all out)

https://github.com/HelloFax/hubot-auth-middleware (should not that the npm package has ext on the end of it, the directions are not accurate on the package name)

the external-scripts.json would then look like this:
```
[
  "hubot-stackstorm",
  "hubot-auth",
  "hubot-auth-middleware-ext",
  "hubot-help"
]
```

obviously installing hubot-stackstorm from this repo instead of npm.


emedvedev:
Incredible! Awesome work.

I'll try to find the time to review this asap, and figure out a way to include it in the st2chatops package as well.

silverbp:
@emedvedev I noticed on the original code there was support for sending a signal to reload the commands and then there's also a timer that reloads them by default every 120 seconds, does something in the st2 stack trigger the chatops service to reload commands and/or do we need support for reloading commands both ways?

package.json
```diff
@@ -1,12 +1,13 @@
 {
-  "name": "hubot-stackstorm",
-  "description": "A hubot plugin for integrating with StackStorm event-driven infrastructure automation platform.",
-  "version": "0.4.5",
-  "author": "StackStorm, Inc. <info@stackstorm.com>",
+  "name": "hubot-stackstorm-auth",
```
lakshmi-kannan: What is going on in this file? :)

Did you intend to open this PR here or your fork?

silverbp: I have this automated to setup our stackstorm box and needed a different package name in npm. I can branch and change that back for the merge. Wasn't sure when this was going to get more attention.

blag:
@silverbp Both ways of (re) loading commands will need to be supported in this rewrite.

I like this, but can you rebase on current `master`?

blag:
I have refactored the chat provider adapters in #186 using part of this PR as a blueprint. Hopefully this makes it easier for the community to add chat provider adapters for other platforms.

Hopefully we can refactor `scripts/stackstorm.js` (which is now [`src/stackstorm.js`](https://github.com/StackStorm/hubot-stackstorm/blob/master/src/stackstorm.js)) soon.

@silverbp Thank you for taking the time to write this PR, I have learned a great deal about Javascript coding practices from reading your code.

Although we won't be using this PR directly when refactoring `src/stackstorm.js`, I will keep this PR open until we do.

blag:
The `scripts/stackstorm.js` module was moved, split up, and partially refactored in #187.

## Further Work ##

Now that `StackStormApi` has been turned into a JS class, it should be easier to write unit tests for each of its public methods, and to update it to use new style JS classes with the `class` and `extends` keywords.

* [ ] Change the `stop()` method to shutdown the previously opened `st2stream` instead of opening a new one and immediately shutting it down - requires manual testing to make sure it doesn't break anything, or add a comment that `st2client.js` caches the connection and reuses it, so opening the same connection simply reuses the old connection and correctly shuts it down
* [ ] Add unit tests for `StackStormApi` public methods
* [ ] Investigate moving things into the `stackstorm.js` wrapper
  - [ ] the `process.on` unhandled rejection handler registration
  - [ ] the `robot.error` handler registration
  - [ ] the `robot.respond` handler registration
  - [ ] the `robot.router.post` endpoint (see [`stackstorm.js:51`](https://github.com/StackStorm/hubot-stackstorm/blob/96d6d3b4dedf056791704f561378c40a185a0ab0/src/stackstorm.js#L51) from this PR)
  - [ ] any other endpoints (I expect at least one to be added for ChatOps authentication as part of ChatOps RBAC)
* [ ] Decouple `stackstorm_api.js` from `stackstorm.js` a bit, similar to how it is implemented here
  - [ ] have `StackStormApi` inherit from `EventEmitter`
  - [ ] emit [`st2.chatops_announcement`](https://github.com/StackStorm/hubot-stackstorm/blob/96d6d3b4dedf056791704f561378c40a185a0ab0/src/lib/stackstorm_api.js#L104) and [`st2.execution_error`](https://github.com/StackStorm/hubot-stackstorm/blob/96d6d3b4dedf056791704f561378c40a185a0ab0/src/lib/stackstorm_api.js#L184) events (consumed in [`stackstorm.js:68`](https://github.com/StackStorm/hubot-stackstorm/blob/96d6d3b4dedf056791704f561378c40a185a0ab0/src/stackstorm.js#L68) and [`stackstorm.js:72`](https://github.com/StackStorm/hubot-stackstorm/blob/96d6d3b4dedf056791704f561378c40a185a0ab0/src/stackstorm.js#L72), respectively)
    + [ ] also possibly emit/handler an event for two-factor *authorization* requests (see `HUBOT_2FA`)
* [ ] Return the promises in `loadCommands` and `start`

blag:
I'm leaving this PR open to track refactoring progress. I'll close it once the refactoring inspired by this PR is complete.

blag:
Marking as stale only so Pull Reminder doesn't keep pinging me about this.

PR #136, opened 2017-05-24
add some support in the spark adapter to send extra dicts

tonybaloney:
Add support for sending file attachments via the extra dict to the spark adapter.


tonybaloney:
@emedvedev  :heart:

emedvedev:
LGTM :heart:

PR #135, opened 2017-03-14
Add rocketchat

cheremushki:
Add support for RocketChat.

PR #138, opened 2017-07-21
Slack post data should use new chat postMessage API

### LINKED ISSUE #121: Problem with @mention in Slack
When `result` message from the bot contains attachment, `@mention` doesn't work in Slack: 
### Good

> ^^ correct & highlighted @mention, you're notified in Slack
### Bad

> ^^ @mention is just a plain text here, no Slack notification from bot

---

As @emedvedev said, something was changed in Slack API.

Original discussion with code example: 
- https://github.com/armab/st2-chatops-aliases/pull/2

lakshmi-kannan:
Based on https://github.com/slackapi/hubot-slack/pull/355/files#diff-04c6e90faac2675aa89e2176d2eec7d8R86 and https://api.slack.com/methods/chat.postMessage#formatting

Also closes #121

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+0.6.0
+-----
+* Update Slack to use new chat postMessage API from 'hubot-slack' v4
+
 0.5.1
 -----
 * Update uuid to version 3.0.0
--- a/lib/post_data.js
+++ b/lib/post_data.js
@@ -32,12 +32,21 @@ function SlackDataPostHandler(robot, formatter) {
 SlackDataPostHandler.prototype.postData = function(data) {
   var recipient, attachment_color, split_message,
       attachment, pretext = "";
+  var envelope;
 
   if (data.whisper && data.user) {
     recipient = data.user;
+    envelope = {
+      "user": data.user
+    };
   } else {
     recipient = data.channel;
     pretext = (data.user && !data.whisper) ? util.format('@%s: ', data.user) : "";
+    envelope = {
+      "room": data.channel,
+      "id": data.channel,
+      "user": data.user,
+    };
   }
 
   if (data.extra && data.extra.color) {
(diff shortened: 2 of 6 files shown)


lakshmi-kannan:
Tested these changes in a hubot instance I spun up.

jszaszvari:
This would break compatibility with all the "slack clones" out there as that API is very proprietary to just slack. 

There is probably 5-6 other solutions out there that are able to use hubot-stackstorm at the moment due to them being compatible with the old method.

armab:
Bumping this version to  `0.5.1` -> `0.6.0` then, to reflect this change.
FYI it's needed because we switch to `hubot-slack` 4+ here https://github.com/StackStorm/st2chatops/pull/78

jszaszvari:
Cool in the mean time everyone who doesn't use slack but uses a platform that's not compatible with slacks proprietary API will update and have all of their chat ops infrastructure stop working.

armab:
@jszaszvari can you clarify in more details how?
I didn't work on this, but my understanding is that the change affects SlackDataPostHandler only which is responsible for Slack and so other adapters should just use their own handlers + custom hubot plugins.

jszaszvari:
@armab Theres been a custom handler here for nearly 5 months that has not been looked at thats for one of the largest Slack competitors - I know of at least 6 different companies using this instead of slack and it just works out of the box without any modifications needed.

https://github.com/StackStorm/hubot-stackstorm/pull/135#issuecomment-315594543

I'm reading through the changes now, I might be mistaken on how it works - So Im trying to test, because when I apply this change, messages just stop coming through completely .

If the PR above could be looked at I know thats going to stop a lot of pain for a lot of people - I've been testing it in production now for a while and it works perfect. 

Just trying to figure  out why if i apply the changes that are coming out it all stops working.

armab:
@jszaszvari Are you referring to any specific Hubot adapter?
Are you saying you're using `hubot-slack` adapter for Rocketchat? If so, I think the right way is to use Rocketchat with `hubot-rocketchat` hubot adapter.

Will it help if I'll merge + release https://github.com/StackStorm/hubot-stackstorm/pull/135 first?

armab:
As figured during the testing, this also fixes https://github.com/StackStorm/hubot-stackstorm/issues/121

PR #141, opened 2017-08-21
Feature/add error message cmd not found

mbrannigan:
If someone gives the chatbot a command that is not registered, provide an error message back to the user.


mbrannigan:
I'm not 100% happy with the approach for ignoring `help`.  Is there some registry of valid yet unregistered commands that the command could be compared against?

mbrannigan:
Closing - duplicate of https://github.com/StackStorm/hubot-stackstorm/pull/109

Issue #143, opened 2017-10-07
Update st2client.js to support customizing st2stream url


shusugmt:
Currently, URL for st2stream is fixed to `$ST2_API/stream`

This is not ideal if running 1ppc mode in st2-docker because without that, st2chatops need to rely on st2web container running nginx which proxies `/api/stream` to st2stream backend.

armab:
:+1: for a configurable env var, similar to existing [`ST2_AUTH_URL`](https://github.com/StackStorm/st2chatops/blob/9de21f73beba4d0ff1e106958a80f5d33a2053ef/st2chatops.env#L26).

shusugmt:
@armab I just opened PR for this. Can you please check it?

shusugmt:
#144 solved this.

PR #144, opened 2017-11-06
Add ability to specify stream URL

shusugmt:
Solves #143 

- `st2client.js` needs to be updated since 0.4.3 does not support specifying stream url
- Use `utils.parseUrl()` just like `ST2_AUTH_URL` do, even the same logic is implemented in `st2client.js` ([here](https://github.com/StackStorm/st2client.js/blob/master/index.js#L50) and [here](https://github.com/StackStorm/st2client.js/blob/master/index.js#L63))
    - We can just pass single string to `opts.auth` or `opts.stream` and let `st2client.js` handle URL parsing, but that is out of the scope here. If we want to do that, we should do it in separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 
 node_js:
-  - "0.12"
   - "4"
   - "5"
   - "6"
--- a/scripts/stackstorm.js
+++ b/scripts/stackstorm.js
@@ -66,6 +66,9 @@ env.ST2_SLACK_FAIL_COLOR = env.ST2_SLACK_FAIL_COLOR || 'danger';
 // Optional, if not provided, we infer it from the API URL
 env.ST2_AUTH_URL = env.ST2_AUTH_URL || null;
 
+// Optional, if not provided, we infer it from the API URL
+env.ST2_STREAM_URL = env.ST2_STREAM_URL || null;
+
 // Command reload interval in seconds
 env.ST2_COMMANDS_RELOAD_INTERVAL = parseInt(env.ST2_COMMANDS_RELOAD_INTERVAL || 120, 10);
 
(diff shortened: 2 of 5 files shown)


LindsayHill:
@armab I think that dropping nodejs 0.12 tests makes sense. (https://github.com/StackStorm/hubot-stackstorm/pull/144/commits/d6092198162bc008b059adb5a3fa33b69296a64c). What do you think?

Kami:
Change looks good to me :+1: 

/cc @armab

scripts/stackstorm.js
```diff
@@ -66,6 +66,9 @@ env.ST2_SLACK_FAIL_COLOR = env.ST2_SLACK_FAIL_COLOR || 'danger';
 // Optional, if not provided, we infer it from the API URL
 env.ST2_AUTH_URL = env.ST2_AUTH_URL || null;
 
+// Optional, if not provided, we infer it from the API URL
+env.ST2_STREAM_URL = env.ST2_STREAM_URL || null;
```
armab: Let's include the corresponding description in `README.md`: https://github.com/StackStorm/hubot-stackstorm/blob/master/README.md#plugin-environment-variable-options

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 
 node_js:
-  - "0.12"
```
armab: Regarding dropping `node.js 0.12`, needs also a change here: https://github.com/StackStorm/hubot-stackstorm/blob/master/package.json#L56-L58

shusugmt:
@armab Thanks for review.

> CHANGELOG update mentioning:
> - dropping node.js 0.12 support
> - updating st2client to 1.1.1
> - adding ST2_STREAM_URL

I thought this should be done in release commit?
- https://github.com/StackStorm/hubot-stackstorm/commit/af65d339e08aa6d7959a0c5fa3720ebe705dc188
- https://github.com/StackStorm/hubot-stackstorm/commit/751f2bec4dceb53e2e16c536feb7e26a77d70359

armab:
The previous changes went into Changelog in one commit because there was nothing else to add (1 change = 1 release). In case of more changes, - it's preferred to include what was done in the changelog as part of PR. Would be easy for release manager so he don't need to grab the git history.

This is similar to what we do in StackStorm/st2, ex: https://github.com/StackStorm/st2/pull/3849/files

shusugmt:
@armab changelog updated!

armab:
Merged, Thanks!

PR #146, opened 2017-11-07
Enable Hubot hear" listeners in StackStorm alias configurations

ahubl-mz:
# Overview #

This PR uses Hubot's `hear()` function to listen in on channels and match against regexes dynamically loaded from StackStorm.

#### An action alias that can match multiple times in a message ####

```yaml
---
name: "jira_get_issue"
pack: "jira"
action_ref: "jira.get_issue"
description: "Get a blurb from a Jira issue."
formats:
  - display: "jira issue <issue key>"
    representation:
      - "jira i(?:ssue)? (?P<issue_key>[A-Z][A-Z0-9]+-[0-9]+)"
  - display: "<issue key>"
    representation:
      - (?:^|\b)(?P<issue_key>[A-Z][A-Z0-9]+-[0-9]+)(?:\b|$)
    regex_flags: g
```

This action alias allows hubot to match a Jira issue key multiple times in one message, so the following Slack message would dispatch multiple times to StackStorm:

> Hey, I started working on SERVICEDESK-1234, but I got blocked by INFRASTRUCTURE-4321. I switch over to working on SERVICEDESK-1235 instead.

Specifically, with this change, Hubot would dispatch three different times to StackStorm:

1. Once for the issue key `SERVICEDESK-1234`
2. Once for the issue key `INFRASTRUCTURE-4321`
3. Once for the issue key `SERVICEDESK-1235`

The `regex_flags` key is exposed to hubot-stackstorm through StackStorm's API, so H-S can compile the regular expression with the appropriate flags.

## Extensibility ##

Our current actions simply respond with a blob of the the issue description in Slack, saving our users a round trip to Jira (which could itself include a login) to read the issue. However, the exact action is not limited to this.

I changed the regex engine from Node.js' default regex engine to the XRegExp engine. I did this because it's compatible with both Python group captures so only one regex format needs to be written.

# Motivation #

This PR sets the stage for some of our changes to the [StackStorm Jira pack](https://github.com/StackStorm-Exchange/stackstorm-jira). We are working on upstreaming those changes as well, once these changes are put into place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+
+*.sublime-project
+*.sublime-workspace
+
 node_modules/
 
 npm-debug.log
--- a/lib/command_factory.js
+++ b/lib/command_factory.js
@@ -19,6 +19,7 @@ limitations under the License.
 
 var _ = require('lodash');
 var utils = require('./utils.js');
+var XRegExp = require('xregexp');
 
 
 /**
@@ -42,7 +43,7 @@ function CommandFactory(robot) {
 }
 
 CommandFactory.prototype.getRegexForFormatString = function(format) {
-  var extra_params, regex_str, regex;
+  var extra_params, regex, regex_str = format;
 
   // Note: We replace format parameters with ([\s\S]+?) and allow arbitrary
   // number of key=value pairs at the end of the string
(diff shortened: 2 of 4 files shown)


LindsayHill:
Thanks @ahubl-mz. Just FYI, I edited the link location for that Jira pack to point to the new ST2 Exchange location.

ahubl-mz:
I think I submitted this PR by accident 😆 - this is a duplicate of PR #148.

@lhill I have ported your correction to #148. Thanks!

Sorry for the noise.

PR #147, opened 2017-11-08
Break up long messages multiple ways to avoid Slack API limits

ahubl-mz:
# Overview #

This PR adds support for sending multiple attachments in StackStorm actions, allows bots to impersonate a username by sending the `username` and `icon_emoji` top-level keys, and has hubot-stackstorm go to great lengths to send Slack messages without erroring out.

# Examples #

#### The old way to send Slack attachments from StackStorm actions ####

```yaml
result:
  format: ...
  extra:
    slack:
      author_name: Jira_Bot
      author_link: "https://stackstorm.com"
      author_icon: "https://stackstorm.com/favicon.ico"
      color: "#042A60"
      fallback: "Info about Jira ticket {{ execution.result.result.key }}"
      title: "{{ execution.result.result.key }}"
      title_link: "{{ execution.result.result.url }}"
      fields:
        -
          title: Summary
          value: "{{ execution.result.result.summary }}"
          short: false
```

would be serialized to the following Slack JSON:

```json
{
  "attachments": [
    {
      "author_name": "Jira Bot",
      "author_link": "https://stackstorm.com",
      "author_icon": "https://stackstorm.com/favicon.ico",
      "color": "#042A60",
      "fallback": "Info about Jira ticket {{ execution.result.result.key }}",
      "title": "{{ execution.result.result.key }}",
      "title_link": "{{ execution.result.result.url }}",
      "fields": [
        {
          "title": "Summary",
          "value": "{{ execution.result.result.summary }}",
          "short": false
        }
      ]
    }
  ]
}
```

Note how there is no way to send multiple Slack attachments or impersonate a user with top-level keys.

#### The new way to send attachments and top-level keys from StackStorm ####

```yaml
result:
  format: ...
  extra:
    slack:
      icon_emoji: ":jira:"
      username: Jira Bot
      attachments:
        -
          fallback: "Info about Jira ticket {{ execution.result.result.key }}"
          color: "#042A60"
          title: "{{ execution.result.result.key }}"
          title_link: "{{ execution.result.result.url }}"
           fields:
            -
              title: Summary
              value: "{{ execution.result.result.summary }}"
              short: false
```

would be serialized to the following Slack JSON:

```json
{
  "icon_emoji": ":jira:",
  "username": "Jira Bot",
  "attachments": [
    {
      "fallback": "Info about Jira ticket {{ execution.result.result.key }}",
      "color": "#042A60",
      "title": "{{ execution.result.result.key }}",
      "title_link": "{{ execution.result.result.url }}",
      "fields": [
        {
          "title": "Summary",
          "value": "{{ execution.result.result.summary }}",
          "short": false
        }
      ]
    }
  ]
}
```

In this second example, StackStorm pack authors can send multiple attachments, and they have full access to top-level keys in the Slack message JSON, allowing them to impersonate users via the `username` and `icon_emoji` keys.

# Reverse Compatibility #

We remain 100% reverse compatible with previously written StackStorm pack actions (see the changes in `lib/post_data.js`. And although we maintain this compatibility, we would like to deprecate the older style of Slack messages in favor of our newer style.

# Additional Considerations #

I created a new module - `lib/messages.js` - to handle sending messages to Slack. It exposes a single function `buildMessage`, but has many internal functions for "chunking" up different parts of a Slack message.

I also added a few new utility functions.

# Tests #

The tests for the `lib/messages.js` module are all contained in `tests/test-messages.js`. While they all call the `buildMessage` function, each test message is structured to tests different codepaths within the `lib/messages.js` module.

I have added tests for the utility functions (as well as fixing a typo) in `tests/test-utils.js`, and added tests in `tests/test-postdata.js` for the new code I added to `lib/post_data.js`.

#### Coverage ####

There should be 100% test coverage for all additional code. I'm juggling branches and commits right now trying to create multiple PRs, but I'm happy to post results from my coverage tool if you would like.

# Documentation #

I have not yet had time to document the extra Slack formatting in StackStorm, but I'm more than happy to create a PR to the st2docs to document the new action format and the older deprecated one.

# Motivation #

Some actions in our packs send multiple attachments to Slack. Others impersonate another bot user (JiraBot) by sending the `username` and `icon_emoji` keys in the message JSON. Both of these features are now supported by hubot-stackstorm.

Additionally, StackStorm hands off messages to hubot-stackstorm to send to Slack, but there isn't a robust error checking mechanism to report an error state back to the StackStorm/Brocade status page. The Slack API docs say that no message - including attachments - should be larger than 4KB [see the bottom of the (Message Formatting section](https://api.slack.com/docs/message-formatting#message_formatting)). However, there is no size checking performed by hubot-stackstorm, so if it is handed a message that is too large, Slack may not accept the message and return an error, and the error may go unreported, except in an error log somewhere that nobody reads. Since the `execution.result.result.summary` field in the action may be longer than 4000 characters, there is no way for pack authors to guarantee that their messages will be smaller than Slack's maximum message size requirements.

Instead of waiting for an error when sending through Slack, we attempt to guess whether or not a message is too large, and if so, we break the message into chunks to send to Slack using various methods. This won't prevent all errors, but it will prevent one of the most common types of errors: message too large. Additionally, if we cannot safely send a message chunk to Slack, we send a small error message to Slack instead. This should inform users that an error happened and enable them to report it to their StackStorm or Slack administrator.

# Final Notes #

I'm happy to answer any questions or respond to any comments, criticisms, or discussion. Thanks!

# Edits #

* Added link to Slack documentation about message size limits.


ahubl-mz:
Adding a polyfill for `Object.assign`...

ahubl-mz:
Polyfill added - tests now pass.

lib/messages.js
```diff
@@ -0,0 +1,254 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+"use strict";
+
+var utils = require('./utils.js');
+
+// Polyfill from:
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+if (typeof Object.assign != 'function') {
```
lakshmi-kannan: Sad you had to port this.

lib/messages.js
```diff
@@ -0,0 +1,254 @@
+/*
```
lakshmi-kannan: All the code here is specific to slack. Should we rename this to lib/slack-messages.js?

lakshmi-kannan:
Really good PR. Something that'd be good is to test this on a box with these changes included. Have you already tested that? I see that the code is backward compatible but I want to make sure we don't break the chatops aliases we ship as part of core st2 (like https://github.com/StackStorm/st2/blob/master/contrib/packs/aliases/pack_get.yaml). 

In another note, it would be good to see what it takes to migrate to latest ECMAScript. Last time I looked at it, it didn't seem like a simple change but I am not a js expert either. I hope migrating to latest ECMAScript will help us avoid introducing code like Object.assign. 

Thanks for the test cases. Those definitely help when reading complicated code like this. Can you post once you validate core aliases work?

lib/messages.js
ahubl-mz: ¯\_(ツ)_/¯

lib/messages.js
ahubl-mz: Derp, yes, that's a good idea.

ahubl-mz:
I know that this was running on our internal dev instance, but I haven't checked the core chatops aliases. I'll do that and report back.

Supporting the latest ECMAScript won't really change what polyfills we have to do. To minimize the amount of polyfills we will need to drop support for older JS interpreters/environments.

Personally\*, I think dropping support for Node.js < v4 and io.js would go a long way towards that end, but it really boils down to what you main StackStorm devs think.

\* Please note that the opinions expressed are my own and do not necessarily reflect those of my employer.

lib/messages.js
ahubl-mz: Done, Travis tests are running.

ahubl-mz:
@lakshmi-kannan I confirmed that the aliases in the `core` pack all work fine with this.

ahubl-mz:
According to @Kami in my [other PR](https://github.com/StackStorm/hubot-stackstorm/pull/148#issuecomment-343908079), StackStorm is heading to more of a "thin" client model.

If that's also the case here, I can put this logic into [stackstorm-slack](https://github.com/StackStorm-Exchange/stackstorm-slack) instead.

lakshmi-kannan:
@ahubl-mz Glad to know the core chatops aliases work well. 

On "thin client" model, I think the parsing of chatops commands and matching them is the thing that has been implemented server side. Your code deals with the reverse path which is I have the execution results and I want to post to slack. This piece is not implemented server side and I don't think it would be in the near future. If we take the approach, we have to write specific chatops adapters for each client which is already enabled by things like hubot. I could be wrong here and I'd be glad to be corrected. So I think this PR is still reasonable to be merged. 

Porting these changes to stackstorm-slack would be really really appreciated. Many users consume the slack pack and the features you implemented would be useful for them too.

ahubl-mz:
I'll see what I can do about porting it to `stackstorm-slack`, but it may not be a high priority if it gets merged in here (but please do merge it in here 😆).

Kami:
@ahubl-mz @lakshmi-kannan 

Yes, only matching is handled server side.

Formatting of the result appropriately for the medium in question (slack / irc / twitter) should still be the responsibility of a bot / client adapter.

Having said that, support for multiple attachments, etc. in slack pack would be great as well :)

ahubl-mz:
Actually, thinking about this a little bit more, the whole reason we go through all of this trouble in `hubot-stackstorm` is because we don't have a good way to report errors once ST2 hands off the message to send to Slack to H-S. Autonomously breaking up messages, trying to preserve their original formatting as close as possible, and each piece off individually is our last resort before erroring *forward* to Slack.

But in the Slack pack, we can detect that a message is too long and easily report that condition as an error back to StackStorm, using StackStorm's normal error flow, which shows up in the BWC web UI, etc.

So checking message lengths in the Slack pack and "eagerly" erroring back to ST2 is probably a good idea, but I think that trying to autonomously breakup messages and send each piece like I do here isn't the best idea in that case. If we can that a message will be too long and cause a Slack error, we should either:

1. Refuse to even send the message to begin with, treat the too-long message as a condition that caused an error, and send the error back to ST2 (eg: "error back").
2. Knowing that H-S will break it up by itself, we can simply ship it off and rely on the error forwarding in H-S to dump the error message into Slack (eg: "error forward").

Note that in #2 we can still try to detect too-long  report a warning back within StackStorm about the message length.

@lakshmi-kannan, @Kami Your thoughts?

armab:
@ahubl-mz or @lakshmi-kannan Please include the missing CHANGELOG entry in https://github.com/StackStorm/hubot-stackstorm/blob/master/CHANGELOG.rst

Also I see the change is tagged with `v0.8.0`. @Kami are we going to release the new `hubot-stackstorm` ver with this change or there are more things left/planned?

PR #148, opened 2017-11-08
Hubot hear and multiple dispatch

ahubl-mz:
# Overview #

This PR uses Hubot's `hear()` function to listen in on channels and match against regexes dynamically loaded from StackStorm.

# Example #

#### An action alias that can match multiple times in a message ####

```yaml
---
name: "jira_get_issue"
pack: "jira"
action_ref: "jira.get_issue"
description: "Get a blurb from a Jira issue."
formats:
  - display: "jira issue <issue key>"
    representation:
      - "jira i(?:ssue)? (?P<issue_key>[A-Z][A-Z0-9]+-[0-9]+)"
  - display: "<issue key>"
    representation:
      - (?:^|\b)(?P<issue_key>[A-Z][A-Z0-9]+-[0-9]+)(?:\b|$)
    # Pass "global" flag to XRegExp so regex can match multiple times per message
    regex_flags: g
```

This action alias allows hubot to match a Jira issue key multiple times in one message, so the following Slack message would dispatch multiple times to StackStorm:

> Hey, I started working on SERVICEDESK-1234, but I got blocked by INFRASTRUCTURE-4321. I'm switching over to working on SERVICEDESK-1235 instead.

Specifically, with this change, Hubot would dispatch three different times to StackStorm:

1. Once for the issue key `SERVICEDESK-1234`
2. Once for the issue key `INFRASTRUCTURE-4321`
3. Once for the issue key `SERVICEDESK-1235`

The `regex_flags` key is exposed to hubot-stackstorm through StackStorm's API, so H-S can compile the regular expression with the appropriate flags.

# Extensibility #

Our current actions simply respond with a blob of the the issue description in Slack, saving our users a round trip to Jira (which could itself include a login) to read the issue. However, the exact action is not limited to this.

Because StackStorm exposes the YAML root keys over its API, hubot-stackstorm can pick up keys that were not initially exposed, such as the `regex_flags` key. This key allows pack developers to specify the flags with which the representation regex should be compiled. By passing the `g` flag, the regex is compiled to match the message multiple times, and each match then gets dispatched to StackStorm. Other regex flags can also be passed to XRegExp using this mechanism.

Finally, with all of that plumbing in place, I have H-S pull the action aliases and use Hubot's `hear()` function so chat bots can respond to appropriate messages even if they were not @mentioned directly.

# Additional Considerations #

I changed the regex engine from Node.js' default regex engine to the XRegExp engine. The original Javascript regex engine does not support named group captures at all, but Python's standard `re` module does. When pack authors specify named groups using Python's extension syntax (eg: `(?P<group_name>...)`), the native Javascript regex engine barfs on the Python-specific named group capture. The XRegExp package properly interprets the Python named capture group syntax, which allows named capture groups in the compiled regex. XRegExp is a wrapper around the built-in Javascript regex engine (it does a little bookkeeping to pair up a numbered capture group with a named capture group), so its both backwards compatible from a matching/calling perspective and from a performance perspective. By using XRegExp, we avoid having to remove or convert Python's named capture group syntax, so the Python-compatible regex that pack authors are already comfortable writing can also automatically be used in Hubot.

# Tests #

I have added tests for `lib/command_factory.js`, but I found it difficult to add automated tests `scripts/stackstorm.js`. I have tested it manually by deploying this code to our dev StackStorm server and running messages into it via Slack.

If you have any testing suggestions for me I am very interested.

#### Coverage ####

According to the coverage report, the tests I added to `tests/test-command-factory.js` have covered 100% of the code I changed in `lib/command_factory.js`.

# Motivation #

This PR sets the stage for some of our changes to the [StackStorm Jira pack](https://github.com/StackStorm-Exchange/stackstorm-jira). We are working on upstreaming those changes as well, once these changes are put into place.

Our final intent is to have a JiraBot listen in on the channel, and when anybody mentions a Jira issue (without necessarily @-mentioning the bot directly), the JiraBot pulls the issue summary from Jira and dumps it into the channel (with pretty formatting, of course, once #147 is merged in).

# Final Notes #

This is some of the first in-depth Javascript I have written, so if I am not following coding conventions, or if there are better ways to implement things, please let me know.

Also, I have not tested this on anything other than Mac OS X with Node.js v8.4.0.

As always, I'm happy to answer any questions or respond to any comments, criticisms, or discussion. Thanks!

### DIFF
--- a/lib/command_factory.js
+++ b/lib/command_factory.js
@@ -24,6 +24,23 @@ var utils = require('./utils.js');
 var XRegExp = require('xregexp');
 
 
+// Polyfill from:
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes
+if (!String.prototype.includes) {
+  String.prototype.includes = function(search, start) {
+    if (typeof start !== 'number') {
+      start = 0;
+    }
+
+    if (start + search.length > this.length) {
+      return false;
+    } else {
+      return this.indexOf(search, start) !== -1;
+    }
+  };
+}
+
+
 /**
  * Manage command for stackstorm, storing them in robot and providing
  * lookups given command literals.
--- a/lib/command_factory.js
+++ b/lib/command_factory.js
@@ -19,6 +19,7 @@ limitations under the License.
 
 var _ = require('lodash');
 var utils = require('./utils.js');
+var XRegExp = require('xregexp');
 
 
 /**
@@ -39,10 +40,13 @@ function CommandFactory(robot) {
 
   // Maps command format to a compiled regex for that format
   this.st2_commands_regex_map = {};
+
+  // Maps command format to a compiled regex for that format
+  this.st2_commands_global_regex_map = {};
 }
 
-CommandFactory.prototype.getRegexForFormatString = function(format) {
-  var extra_params, regex_str, regex;
+CommandFactory.prototype.getRegexForFormatString = function(format, flags) {
+  var extra_params, regex, regex_str = format;
 
   // Note: We replace format parameters with ([\s\S]+?) and allow arbitrary
   // number of key=value pairs at the end of the string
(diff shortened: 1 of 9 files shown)


ahubl-mz:
Working on fixing these failures today.

ahubl-mz:
Added polyfils for Node.js v0.12 and io.js.

LindsayHill:
Since we now install Node.js v6.0 by default, maybe we should just update our travis config here, and drop older Node.js versions, and io.js?

Save others some pain with dealing with this kind of fix.

Kami:
First thank you for this contribution :+1: 

Now to add some context and figure out how we can integrate this change.

A while ago we have started to move away from the "fat / smart client" approach (regexes and other logic is encoded in the client) to moving all the alias matching and executing functionality to the server side (API). Client / adapter should only be responsible for authenticating and formatting messages for the medium in question and nothing else.

This approach has many benefits, but the most important ones is that the clients are very simply (dumb) wrappers around APIs which makes them much easier (and cheaper in terms of resources) to maintain and it also makes it much easier and faster to develop adapters for other chat systems.

The server side change has already been made (https://github.com/StackStorm/st2/pull/3773) and client side (hubot) change was on our TODO in the near future.

In short - with the new approach, the client will send every chat line to the new "match and execute" StackStorm API and listen for the response. We are aware of the potential performance overhead of this approach, but it's still much better than current approach which is very hard and expensive to maintain, tests, etc. (we basically have duplicated / same logic on the client and server).

And as far as the performance overhead goes, it's not bad unless you have 100s of messages per-second in your channel, but in that case you should probably invite bot to a different channel with less traffic and existing approach would be slow in such scenario as well.

We should figure out how and if we can integrate those changes in the new future approach.

ahubl-mz:
@LindsayHill I'd prefer to adjust the Travis config in a different PR, just to keep the scope of this PR more narrow.

lib/command_factory.js
```diff
@@ -39,26 +59,72 @@ function CommandFactory(robot) {
...
+    regex_str = regex_str.replace(/(\s*){{\s*(\S+)\s*=\s*(?:({.+?}|.+?))\s*}}(\s*)/g, '\\s*($1(?<$2>[\\s\\S]+?)$4)?\\s*');
+
+    // {{ option }} -> ([\s\S]+?)
+    // Grab the whitespace on either side so regexes can specify whether or not the whitespace is optional
+    // regex_str = regex_str.replace(/(\s*){{\s*([\w_]+).*?\s*}}(\s*)/g, '$1(?<$2>[\\s\\S]+?)$3');
+    var parameter_re = new XRegExp('(?<before_ws>\\s*){{\\s*(?<var>[\\w_]+).*?\\s*}}(?<after_ws>\\s*)', 'g');
+    regex_str = XRegExp.replace(regex_str, parameter_re, '${before_ws}(?<${var}>[\\s\\S]+?)${after_ws}');
+
+    // Only anchor if we aren't anchored already
+    if (regex_str.charAt(0) !== '^') {
```
ahubl-mz: I don't like how I do this, but I don't know of a way to parse Javascript's regex objects like how I did in [PR #3789](https://github.com/StackStorm/st2/pull/3789). I'm open to suggestions.

ahubl-mz:
@Kami I didn't realize that API endpoint existed - I'll re-implement this in ST2 proper (I like coding in Python more anyway).

The way this PR implements matching is:

```
loop through all non-global alias regexes
    if one matches -> dispatch to ST2 -> stop searching for additional aliases
    if none match ->
        loop through all global alias regexes
            if one matches ->
                loop through all matches of that alias -> dispatch each match to ST2
                -> stop searching for additional aliases
        if none match -> error out
```

Do those semantics sound reasonable to implement in ST2? I can go into more detail about our motivations if necessary.

Kami:
@ahubl-mz Yeah that sounds reasonable to me.

And yeah, since the match will now happen server-side we should try to be careful and make it as efficient and fast as possible (ideally, complexity should be o(o) where n is number of aliases in the system).

ahubl-mz:
I've taken a look at reimplementing this within StackStorm, and I need some guidance.

In `ActionAliasExecutionController.match_and_execute` (see `st2api/controllers/v1/aliasexecution.py`), it returns the result of `self.post(...)` on line 93. The `post()` function returns a `Response` object with a status and a dictionary of result data.

This is problematic because I will need to call `post()` multiple times in `match_and_execute`, and there's no good way to maintain reverse compatibility with the existing API that way.

So my options are:

1. Modify `match_and_execute()` to return a single result with a list of dictionaries of result data from the (potentially) multiple executions, breaking the current API.
2. Add another endpoint (probably something like `multiple_match_and_execute`) and migrate clients to it.
3. Bump the API version and extend the `match_and_execute()` function as intended without breaking the existing API. Clients can migrate at their leisure.

Which option would you like me to take?

-------------------

Also, about algorithmic complexity - this is another problem. If you take a look at the `match_command_to_alias` function in `st2common/util/actionalias_matching.py`, it matches commands in O(n<sup>2</sup>) time:

```python
```

I think I can work around this by querying the database in a more complex way to only return display dictionaries that contain a `regex_flags` key whose value contains `g` which will avoid the outer loop in that function.

Kami:
@ahubl-mz AFAIK, we don't really have any official consumers of that API endpoint yet, so option 1) is not too bad.

The plan was to add the endpoint first, then migrate the clients (notably hubot-stackstorm) to it in the near future :)

Or perhaps, if we can re-use most of the code and avoid duplication, adding another endpoint "multiple_match_and_execute" works as well.

As far as algorithmic complexity goes - yeah I noticed that while working on it recently, so if you could simplify it, that would be great :)

ahubl-mz:
I will be doing option 1 then, since it (presumably/hopefully) won't break anything major.

My current plan for reducing the algorithmic complexity of the logic is to query Mongo with something like:

```python
cursor = db.inventory.find({"formats": SON([("regex_flags", "g")])}, {"formats.display": 1})
```

That will allow me to remove one of the nested `for` loops, but I'm afraid I might simply be hiding the for loop in the query. However, implementing it that way will allow me to cache the results (if I get that working). ¯\\\_(ツ)\_/¯

Kami:
@ahubl-mz Sounds good and keep us posted and ping us when st2 pr is ready :)

ahubl-mz:
ST2 PR has been submitted for internal legal review. I should have a PR for you by (hopefully early) next week.

ahubl-mz:
Now that StackStorm/st2/pull/3884 is merged, I'll rewrite this to use the new `aliasexecution/match_and_execute` endpoint and submit a different PR.

ahubl-mz:
Okay, so now that the `match_and_execute` endpoint on StackStorm can execute an action multiple times, we can rewrite hubot-stackstorm to use that endpoints exclusively.

That means that we don't have to use `.respond()` anymore, we don't have to query, load, or match against regular expressions in Hubot.

Really, `stackstorm.js` will basically look like:

```js
var sendAck = function (match) {
  // ...
}

forwardToMatchAndExecute(msg) {
  // ...
}

robot.hear(/.*/, function (msg) {
  matches = forwardToMatchAndExecute(msg);

  _.each(matches, sendAck);
});
```

Which is much, much simpler.

Kami:
Yep and I think it should work quite well unless you have a channel with tons of messages per second.

Because of that, it's also important to scope bot hear to a single channel now that every line is sent to the API.

Issue #151, opened 2017-11-30
Support for "Button attachments"


SURAJTHEGREAT:
I am not able to send the message with button attachments. If I send simple message{non-attachment message} its working fine. If i send message with extra param from the stackstorm the message is not getting displayed in the channel.

If I have changed the line 137 in **post_data.js** under **lib** folder from **attachments: content**  to **attachments: content.attachments ? content.attachments : content** then the button gets displayed.

Please help on this issue. Is this a right fix

message from stackstorm
```yaml
action:
  ref: chatops.post_message
  parameters:
    channel: "bq97zw96ypnrjp15h9ey4bhu5o"
    #channel: "#town-square"
    message: "Trying button action"
    extra:
      color: "#ff0000"
      mattermost:
           attachments:
                 - pretext: "This is the attachment pretext."
                   text: "This is the attachment text."
                   actions:
                      - name: "Ephemeral Message"
                        integration:
                              url: "https://x.x.x.x/hubot/buttonaction"
                              context:
                                  action: "do_something_ephemeral"
                      - name: "Update"
                        integration:
                              url: "https://x.x.x.x/hubot/buttonaction"
                              context:
                                  action: "do_something_update"
```

armab:
Without digging into the code, logically your code suggestion looks safe.

Can you open a PR with mentioned changes in `post_data.js`? And someone will review it and hopefully that would be included in hubot-stackstorm.

SURAJTHEGREAT:
@armab : We are facing error **remote: Permission to StackStorm/hubot-stackstorm.git denied to SURAJTHEGREAT.** when trying to git pull the repository. Could you please provide access for the same.

arulvelu21:
@armab : We have created fork method as well to create a pull request. Currently it is showing in pull request under the heading "Button attachement fix for mattermostuser" . But the build Travis CI is getting failed.Please assist.

armab:
Implemented in https://github.com/StackStorm/hubot-stackstorm/pull/152

PR #152, opened 2017-12-01
Button attachement fix for mattermostuser

arulvelu21:
**Based on the issue Support for "Button attachments" #151**
 
Inorder to accept the matteruser adapter message attachments we have changed the attachement json in post_data.js

### DIFF
--- a/lib/post_data.js
+++ b/lib/post_data.js
@@ -132,9 +132,14 @@ MattermostDataPostHandler.prototype.postData = function(data) {
       content.text = chunks[i];
       content.fallback = chunks[i];
 
+      /*
+        Based on the issue Support for "Button attachments" #151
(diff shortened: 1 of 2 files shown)


lib/post_data.js
```diff
@@ -132,9 +132,14 @@ MattermostDataPostHandler.prototype.postData = function(data) {
...
 
+      /*
+        Based on the issue Support for "Button attachments" #151
+        Inorder to accept the matteruser adapter message attachments changed the attachement json
+      */
+
       attachment = {
         room: recipient,
-        attachments: content,
+        content.attachments ? content.attachments : content,
```
armab: It's expected to have a dict here.

I believe the correct construction would be:
```js
attachments: content.attachments ? content.attachments: content,
```

PR #155, opened 2018-01-16
Update to version 0.8.0

lakshmi-kannan:


CHANGELOG.rst
```diff
@@ -1,8 +1,8 @@
 Changelog
 =========
 
-in development
```
armab: May need that string for future `in development` changelog, otherwise contributors less likely to include them

PR #159, opened 2018-10-06
Update components

blag:
Updated dependencies. Not all could be updated at this time - the `hubot-help` package tries to require `coffee-script`, but that was migrated to using the `coffeescript` package instead.

I have another commit that adds coverage results, but I'll wait to push that.


LindsayHill:
> And so I would suggest to go with iterative approach to avoid regressions as much as possible, instead of trying to upgrade everything in one go.

Normally I would agree with you. But we're out of date with so many components that this could take ages. Maybe better to just try upgrading everything, see if it works, or at least can be easily fixed. But then if it turns out to be hard to fix, then go back to iterative changes.

LindsayHill:
Something else we should do as part of this is update node.js too.

blag:
Yeah, I “fixed” tests by upgrading to node 6+. I develop with node 10.

Frankly, I don’t think we need to support node version older than 2 years, like node.js 4 and io.js, but Im keeping an open mind if people have complaints about that. The JS ecosystem moves much faster than we do anyway.

LindsayHill:
Time for us to upgrade node.js anyway, since 6.x is end of support in April 2019. Jumping straight to 10 is the best path. We don't really want to support old versions, it just a bit of a pain for users upgrading since they need to change repos. But if we go to 10.x now, that gives us good coverage for a while.

Last time we bumped version was August 2017.

package.json
```diff
@@ -22,36 +22,48 @@
...
+    "lodash": "^4.17.11",
+    "rsvp": "^4.8.4",
     "st2client": "^1.1.1",
-    "truncate": "^1.0.4",
+    "truncate": "^2.0.1",
     "uuid": "^3.0.0"
   },
   "peerDependencies": {
-    "hubot": "2.x"
+    "hubot": "3.x"
```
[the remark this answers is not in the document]
blag: Yes, good thinking.

CHANGELOG.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 in development
 --------------
 
+0.9.0
+-----
+* Update dependencies (improvement)
+* Upgrade from hubot 2 to 3 (improvement)
```
armab: I think it's more fits the `change` category, rather then improvement.

Also worth including in Changelog other goodies like deprecation of older nodejs versions as a `change` and adding new nodejs versions as supported.

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ Changelog
 in development
 --------------
 
+0.9.0
+-----
+* Update dependencies (change)
+* Upgrade from hubot 2 to 3 (change)
+* Drop support for Node.js < 6 (change)
+* Add support for Node.js >= 6 (change)
```
armab: Isn't `>= 6` too much greedy way of showing it?
ATM we're pinning support to maximum node `10`. When new nodejs version will be released (say `11`), it might be confusing.

I'd propose to be more clear and list explicitly which versions are deprecated and which new versions were added.

blag: If I pin it to a maximum of node 10, and node 11 _doesn't_ break anything, then we will need to revisit this to update the accepted node.js version. Only specifying a minimum version relieves us of a bit of the maintenance burden.

And considering how well we update our node projects, I'm in favor of reducing our maintenance burden.

blag: I'll be more detailed about what dependencies were upgraded though. :)

armab: > If I pin it to a maximum of node 10, and node 11 doesn't break anything, then we will need to revisit this to update the accepted node.js version

By supporting specific versions which we already doing, I mean this: 
https://github.com/StackStorm/hubot-stackstorm/blob/6d8721a596eaaeaa832cccb26099a3a174d1aff4/.travis.yml#L1-L9

If we'll support node 11, - that will come with another `hubot-stackstorm` version update and a new Changelog entry and code change. The support statement shouldn't be that blurred.

CHANGELOG.rst
```diff
@@ -4,6 +4,23 @@ Changelog
...
+* Update gulp plugins (gulp-jshint to 2.1.0, gulp-mocha to 6.0.0, gulp-plumber
+  to 1.2.0) (change)
+* Update to coffee-script 1.12.7 (change)
+* Update to lodash 4.17.11 (change)
+* Update to nock 10.0.0 (change)
+* Update to rsvp 4.8.4 (change)
+* Update to sinon 6.3.5 and sinon-chai 3.2.0 (change)
+* Update to truncate 2.0.1 (change)
+* Update to hubot 3 and hubot-help 0.2.2 (change)
+* Migrate to coffeescript 2.3.2 and coffee-register 2.2.0 (change)
```
armab: Sorry for confusion, I didn’t mean to include in the changelog every minor dependencies that we upgrade.
That looks excessive, look at changelog history.
Major changes are nodejs and hubot which can’t be blurry about exact version diff.

CHANGELOG.rst
blag: It's entirely acceptable (to me) to have differences between "works with", "supported", and "tested" though.

* We can test for more versions than we officially support (I've done this on some projects), because developers who run older versions of dependencies know that they are behind the curve.
* The code may work with more versions than we officially support or test - older or newer versions of node fall into this category.

So yes, once we officially support node 11, we should test with that, add it to the changelog, and officially support it, but if things might still work with node 11, I don't see any reason to explicitly limit the code compatibility just to node 10.

This might be a good time to put the `node` version (which tests the most recent version of node.js at build time) into the `allow_failures` part of the Travis config. That may be a good middle ground?

armab: Yes, code compatibility is good :+1: 
The point is just to be very clear in Changelog/documentation what we support and what's not.

It might be also useful as you'll work on st2chatops: we pin max supported `nodejs` version in both `st2chatops` deb/rpm packages (https://github.com/StackStorm/st2chatops/blob/492b6daaafe4aa7faf83be634e38b54b744ca7d2/debian/control#L14) and `curl|bash` installer: https://github.com/StackStorm/st2-packages/blob/0f2770be1fac382267f9c9cdab27a3b1865ab9f7/scripts/st2bootstrap-deb.sh#L656-L657

LindsayHill: I've been wondering if we should relax our nodejs version dependency in those packages. E.g. support 6/8/10, rather than just 6.

Having a wider range might make things a little easier for folks migrating between versions. It's a pain changing versions due to need to repo changes. Makes upgrades a bit more involved.

armab: +1, setting supported interval of versions would work.
Just hope there won't be any compatibility issues between min (already installed for some users) and max (new installs) version as we update `st2chatops` + `hubot-stackstorm`.

Probably needs another cycle of manual testing?

blag:
Companion PR is StackStorm/st2chatops#116.

LindsayHill:
Don’t forget to update the upgrade section of st2docs, and the install scripts. Will need something similar to last time we bumped node versions

PR #163, opened 2019-03-19
Prefer 'ST2_API_URL' ENV instead of 'ST2_API' for consistency

armab:
> This is spot during the K8s work https://github.com/StackStorm/st2-dockerfiles/pull/19

For consistency with other ENV variables (see https://docs.stackstorm.com/reference/cli.html#configuration), rename `ST2_API` -> `ST2_API_URL`.
The change still supports old `ST2_API` for backwards compatibility if someone is using `hubot-stackstorm` externally.

---

`st2chatops` change to `st2chatops.env` will come next.

### DIFF
--- a/scripts/stackstorm.js
+++ b/scripts/stackstorm.js
@@ -45,7 +45,14 @@ var _ = require('lodash'),
   ;
 
 // Setup the Environment
-env.ST2_API_URL = env.ST2_API_URL || env.ST2_API || 'http://localhost:9101';
+env.ST2_API_URL = env.ST2_API_URL || null;
+if (!env.ST2_API_URL) {
+  if (env.ST2_API) {
+    robot.logger.warning("ST2_API is now deprecated and will be removed in a future release. Instead, please use the ST2_API_URL environment variable.");
+  }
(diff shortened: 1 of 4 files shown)


scripts/stackstorm.js
```diff
@@ -45,7 +45,7 @@ var _ = require('lodash'),
   ;
 
 // Setup the Environment
-env.ST2_API = env.ST2_API || 'http://localhost:9101';
+env.ST2_API_URL = env.ST2_API_URL || env.ST2_API || 'http://localhost:9101';
```
blag: To ensure backwards compatibility, I would do this:

```js
env.ST2_API_URL = env.ST2_API_URL || null;

if (!env.ST2_API_URL) {
  if (env.ST2_API) {
    robot.logger.warning("ST2_API is now deprecated and will be removed in a future release. Instead, please use the ST2_API_URL environment variable (the value can remain the same).");
  }
  env.ST2_API_URL = env.ST2_API || 'http://localhost:9101';
}
```

armab: That's a good addition, works for me :+1:

PR #164, opened 2019-03-26
Cleanup a bit and use hubot-botframework for MS Teams support

blag:
* Adds unit tests for adapters
* Pins hubot-help version to `0.2.2`
* Adds adapter for MS Teams
* Rename `createExecution` function to `sendAliasExecutionRequest` so it's a bit clearer how the function is different than `executeCommand`
* Send `source_context` in `sendAliasExecutionRequest` function
* Adds tests for BotFramework adapter
* Cleans up some (but not all) layering violations with the HipChat adapter
* Adds tests for `normalizeAddressee` functions on adapters
* Adds comments to adapter functions
* st2 changes in StackStorm/st2#4609
* st2chatops in StackStorm/st2chatops#121
* Documentation for configuring Microsoft Teams with BotFramework in StackStorm/st2docs#872


lib/format_data.js
```diff
@@ -88,6 +95,48 @@ MattermostFormatter.prototype.normalizeCommand = function(command) {
...
+  // Remove starting newlines
+  data = data.replace(/^\n/g, '');
+  // Replace single newlines with double newlines
+  data = data.replace(/([^\n])\n([^\n])/g, "$1\n\n$2");
+
+  return data;
+};
+
+MSTeamsFormatter.prototype.formatRecepient = function(recepient) {
+  this.robot.logger.info("Got recipient in formatRecepient: " + JSON.stringify(recepient));
```
Kami: Should probably be debug log level?

lib/post_data.js
```diff
@@ -190,10 +193,54 @@ SlackDataPostHandler.prototype.postData = function(data) {
...
+  }
+
+  // We define a recursive closure that calls itself with the next data to send
+  // after a timeout. This approximates sending synchronous (sequential) HTTP
+  // requests.
+  var sendMessage = function (i) {
+    robot.adapter.send(data.context, messages_to_send[i]);
+
+    if (messages_to_send.length > ++i) {
+      setTimeout(function () { sendMessage(i); }, 300);
```
Kami: Almost like the IRC days where you were limited to how fast you can send the commands :laughing:

Kami: It looks like the "worst" case will be a maximum of two messages?

Kami:
LGTM, looking forward to integration / end to end tests in the future :+1:

lib/post_data.js
blag: Yeah, what's even more annoying is that the underlying JS library takes a `done` callback parameter that is called when the HTTP request resolves, but the `adapter.send` function only accepts a `context` parameter and `messages...` parameter/s. 😆

For now, yes, the worst case will be a `pretext` and a `text` message, but if this adapter gets any more complex (and I expect it to), that will likely change. I completely stole this pattern from the Slack adapter, which is an excellent example of how complex it can get.

lib/format_data.js
blag: Yep, will change.

armab:
Worth also updating CHANGELOG record with this new addition:
https://github.com/StackStorm/hubot-stackstorm/blob/master/CHANGELOG.rst

Issue #77, opened 2015-12-09
Hubot doesn't reconnect to the stream on API restart


emedvedev:
> hm, so after I restart with st2ctl restart, my bot stops receiving events from st2 :simple_smile:
> until I restart the bot too

A community report. I've run into that issue a couple times, too. Not sure if we should solve in st2client or here.

emptywee:
Yes, I reported that the other day. Let me know if you need any info from my end if you can't repro it.

emedvedev:
@emptywee thanks! I can, although not always. This is probably an eventstream module issue: by default a client should attempt reconnects indefinitely, and it looks like the module we use doesn't always do that. Even if it's the case, I think we'll be able to fix it or implement an additional layer of checks on our level.

ticean:
Hi there. I'm running a new installation to trial StackStorm and I *think* this issue biting us fairly hard. Anytime the our Hubot loses connection to the StackStorm API it doesn't attempt reconnect and is then running lame. No more StackStorm goodness. I have to manually restart Hubot :(

Is this on the priority to fix? The functionality I've been able to implement quickly is great! But the reliability here is a big deal.

We're running Hubot independently. I installed this script into our previously existing bot.

Thanks for the help!

dzimine:
Thanks for the report! we will be looking into it, your +1 increments the priority but no committed fix yet.

emedvedev:
@enykeev was looking into it some time ago, but it would require a fairly difficult module rewrite. @enykeev: sup?

emedvedev:
@ticean: did you install with packages or AIO installer? We still have this issue on packages, but AIO should be good. 

In short, this error is caused by the stream consumer module not recognizing error 5xx as a reason for reconnect. In AIO we apply a special fix to give stream errors special treatment: https://github.com/StackStorm/st2workroom/pull/303/files

If you chose packages as your install method, right now you can apply it manually, and in the future we'll hopefully have a better fix.

ticean:
I installed with AIO installer, but I configured this `hubot-stackstorm` plugin into a pre-existing, non-AIO Hubot installation.

As an underlying problem, I'm finding that the stackstorm nginx instance is stopping (or going zombie) every night. I haven't been able to find out what's scheduled to cause that, but it definitely seems like periodic task. The host is dedicated to stackstorm with AIO. If I could find and fix this, then it would definitely lower the urgency of the issue.

[...]

PR #168, opened 2019-05-21
Log errors and exit on authentication issues on Hubot plugin initialization

jinpingh:
Fixes https://github.com/StackStorm/st2chatops/issues/124
1. Capture all authentication related exceptions and exit from `st2chatops`.
2. Create new callback function for `robot.error` to capture uncaught error from hubot.
3. Add unit tests for authentication error.

Closes stackstorm/st2tests#170


scripts/stackstorm.js
```diff
@@ -409,32 +419,36 @@ module.exports = function(robot) {
...
 
     // Install SIGUSR2 handler which reloads the command
     install_sigusr2_handler();
+
+    // Add an interval which tries to re-load the commands
+    commands_load_interval = setInterval(loadCommands.bind(self, false), (env.ST2_COMMANDS_RELOAD_INTERVAL * 1000));
   }
 
-  function stop() {
+  function stop(exitFromProcess) {
```
blag: Since `exitFromProcess` is an options object, we should rename it to something like `opts` or `options`.

scripts/stackstorm.js
```diff
@@ -171,7 +174,11 @@ module.exports = function(robot) {
       .catch(function (err) {
         robot.logger.error('Failed to authenticate: ' + err.message);
 
-        throw err;
+        if (env.EXIT_ON_FAILURES) {
+          stop({exit: true});
```
armab: What if another error is thrown in different place of the code? Will the hubot-stackstorm wait in the same noop state?

If we reached the state when the raw error is thrown to user, - that was an automatic exit before (expected & correct).

I suspect something changed in recent refactorings or NodeJS upgrades?
Some handler was broken?

blag:
> If we talk about the failure which could be potentially recovered by connection retries/or retries with backoff, - we can consider using that instead.

We have this behavior between other microservices of ST2, so it would be nice to have the same behavior in st2chatops. But code-wise, it would be frustrating to keep the semantics of "retry with backoff" in st2chatops up-to-date with how it works in the rest of ST2.

> But if nothing helped, - always exit with error message and non-zero exit code.

This is the easiest thing for us to do. At this point, I think this behavior is a good compromise between trying to behave like the other ST2 services and increased complexity of this codebase/PR.

scripts/stackstorm.js
blag: I was curious about this, so I dug into it. This changed [recently](https://github.com/hubotio/hubot/blob/d2b57bd796749fbb9ca02eff33b7a345bcd40d68/docs/scripting.md#error-handling):

> Previously, an uncaught exceptions would crash your hubot instance. Hubot now includes an uncaughtException handler, which provides hooks for scripts to do something about exceptions.

Hubot catches uncaught exceptions itself now. Previously it would just crash.

armab: Awesome, so that's the root cause.

Sounds like normally we would just need to rely on that handler and add exit with non-zero code without changing any other logic.

@blag Thanks for finding it! :+1:

scripts/stackstorm.js
```diff
@@ -238,6 +239,9 @@ module.exports = function(robot) {
       .catch(function (err) {
         var error_msg = 'Failed to retrieve commands from "%s": %s';
```
blag: Let's also change this to log the full traceback. See https://github.com/hubotio/hubot/blob/master/src/robot.js#L461 for an example.

scripts/stackstorm.js
```diff
@@ -170,8 +170,7 @@ module.exports = function(robot) {
       })
       .catch(function (err) {
         robot.logger.error('Failed to authenticate: ' + err.message);
```
blag: And also change this to log the full traceback.

scripts/stackstorm.js
armab: @jinpingh please pay attention to this ^^ discussion.

Try adding a handler that'll exit on every error/exception instead of patching specific cases and leaving any other possible failure scenarios buggy as reported in StackStorm/st2chatops#124.

scripts/stackstorm.js
```diff
@@ -409,22 +414,28 @@ module.exports = function(robot) {
...
 
-    // Add an interval which tries to re-load the commands
-    commands_load_interval = setInterval(loadCommands.bind(self), (env.ST2_COMMANDS_RELOAD_INTERVAL * 1000));
-
     // Initial command loading
-    loadCommands();
+    loadCommands({exitOnFailure: true});
+
+    // Add an interval which tries to re-load the commands
+    commands_load_interval = setInterval(loadCommands.bind(), (env.ST2_COMMANDS_RELOAD_INTERVAL * 1000));
```
blag: Make sure you bind `loadCommands` with `self`, like it was originally.

scripts/stackstorm.js
jinpingh: @armab Current fixes take care of all st2 related configuration issues.  For `hubot` configuration issue (for examples: With invalid HUBOT_SLACK_TOKEN) it raises exception from `hubot` and  `st2chatops` is still hanging.

blag:
Should close #77 if st2chatops gets automatically restarted, which I think it will in a kubernetes environment.

CHANGELOG.rst
```diff
@@ -8,6 +8,7 @@ in development
 -----
 * Add initial support for Microsoft Teams via BotFramework (new feature)
 * Add unit tests for all adapters (improvement)
+* Allow `st2chatops` to exit from `st2client` failures.
```
armab: Changelog record should belong to likely new `0.9.4` version that'll need also a version bump in `package.json` and Github release + npm publish later.

tests/test-st2-invalid-auth.js
```diff
@@ -0,0 +1,154 @@
...
+  mockedEnv = require('mocked-env'),
+  Robot = require("hubot/src/robot"),
+  Logger = require('./dummy-logger.js');
+
+chai.use(sinonChai);
+chai.use(chaiString);
+
+global.process.exit = sinon.spy();
+
+describe("invalid st2 credential configuration", function() {
```
armab: With these tests in hubot-stackstorm do we still need e2e tests located https://github.com/StackStorm/st2tests/pull/170?

blag: No, we'll close that PR once this PR is merged.

armab:
Thanks @jinpingh and @blag for jumping in!
I'll test the packages from this PR in K8s/Docker environment once again to verify it and will get back to you.

scripts/stackstorm.js
```diff
@@ -370,30 +373,26 @@ module.exports = function(robot) {
...
       exitProcessWithLog('Unable to connect to stream: ', err);
     }).then(function (source) {
       source.onerror = function (err) {
-        // TODO: squeeze a little bit more info out of evensource.js
-        if (err.status === 401) {
-          robot.logger.error('Stream error:', err);
-        } else {
-          var error_message = util.format('stream error: [type: %s; status: %s]', err.type, err.status);
-          exitProcessWithLog(error_message, null);
-        }
```
jinpingh: @blag Sorry for this later comment. Just tested new `stackstorm.js` code and realized by removing this part of code, the user has no idea what is issue when an invalid `ST2_API_KEY` or `ST2_AUTH_TOKEN` is encountered.  Only error message is  `ERROR undefined`. Before this change, it will get more meaningful messages  : 
```
For invalid ST2_AUTH_TOKEN:
[Wed Jun 05 2019 17:31:31 GMT+0000 (Coordinated Universal Time)] ERROR Stream error: Event { type: 'error', status: 401 }
[Wed Jun 05 2019 17:31:31 GMT+0000 (Coordinated Universal Time)] ERROR Failed to retrieve commands from https://localhost/api Unauthorized
[Wed Jun 05 2019 17:31:31 GMT+0000 (Coordinated Universal Time)] ERROR undefined

For invalid ST2_APY_KEY:
[Wed Jun 05 2019 17:38:28 GMT+0000 (Coordinated Universal Time)] ERROR Stream error: Event { type: 'error', status: 401 }
[Wed Jun 05 2019 17:38:28 GMT+0000 (Coordinated Universal Time)] ERROR Failed to retrieve commands from https://localhost/api Unauthorized - ApiKey with 
key_hash=cddbf16c36e281d74efd98b8fc9467d0b68e304ae4b0e0026f9b8ec5a95043fef7393dddd2edb712c4ed499f4752f7e233991b8b9053958748a1be051c30a1a5 not found.
[Wed Jun 05 2019 17:38:28 GMT+0000 (Coordinated Universal Time)] ERROR undefined
```

blag: Are you sure that was where that error message was coming from? Because the `error_message` variable should have been printing something like `stream error: [type: %s; status: %s]`, but the error message was different: `Stream error: Event { type: 'error', status: 401 }`.

jinpingh: Line 394  `util.format('stream error: [type: %s; status: %s]', err.type, err.status);`  printed out `Stream error: Event { type: 'error', status: 401 }`.

scripts/stackstorm.js
armab: OK, I think something related to the discussion happening in this thread + research + was the right path to solve this issue.
We've talked before about global handler that will catch similar unrecoverable issues. I think https://github.com/StackStorm/hubot-stackstorm/pull/172 is the simpler way to solve the root cause.

tests/test-st2-invalid-auth.js
```diff
@@ -0,0 +1,154 @@
...
+    restore_env = mockedEnv({
+      ST2_AUTH_USERNAME: 'nonexistent-st2-auth-username',
+      ST2_AUTH_PASSWORD: 'nonexistent-st2-auth-password'
+    });
+
+    // Load script under test
+    var stackstorm = require("../scripts/stackstorm.js");
+    stackstorm(robot).then(function (stop) {
+      expect(error_spy).to.have.been.calledOnce;
+      expect(error_spy.firstCall.args[0]).include('Environment variables ST2_AUTH_USERNAME, ST2_AUTH_PASSWORD and ST2_AUTH_URL should only be used together.');
```
armab: @blag I think some of these tests would be very useful if extracted as separated PR. They really look nice :+1: 

It feels that we try to put too much into 1 single PR: bugfix, refactoring and significant change of logic. Would be great to stick with several independent and controlled & predictable changes.

blag:
@armab 
> we significantly changed the logic here

Yep.

> and overengineered it

I disagree. The part of this PR that is fixing the issue is the proper way to implement this change according to hubot's documentation on error handling. The code churn that has happened as a result is necessary to ensure that places where connection issues could happen will all cause hubot to shutdown. Additionally, we only wanted to have one way to gracefully shutdown instead of spaghetti code shutting it down in multiple places.

> removing existing self-healing mechanisms

Yep. In a k8s environment, gracefully exiting and relying on k8s or the process manager to restart the process from scratch is a self-healing mechanism. ChatOps is not part of the rest of ST2 microservices, and it has to re-authenticate when it starts up.

> to re-load and recover hubot with `ST2_COMMANDS_RELOAD_INTERVAL` setting.

As long as the connection to StackStorm is fine, this behavior is unchanged: st2chatops will continue to reload commands on a configurable interval. If it cannot connect to StackStorm, the process will die, hopefully be respawned, and the new instance will connect again.

The other option, which is what we originally had, was to make the shutdown/respawn/reconnect behavior configurable via an environment variable, but [you didn't like that solution either](https://github.com/StackStorm/hubot-stackstorm/pull/168#pullrequestreview-241419847), and that's why we refactored it to this behavior.

> I really meant to exit app on unrecoverable failures (when reload stopped to work and hubot was just hanging in noop) by adding some kind of handler for it.

Yes, that's precisely what this PR implements, in a modern, proper manner, following Hubot documentation, including tests, and a minimum amount of refactoring to ensure that errors are not swallowed and that they do propagate up to the shutdown-on-error logic.

armab:
> If it cannot connect to StackStorm, the process will die, hopefully be respawned, and the new instance will connect again.

Previously, if hubot couldn't connect to st2 the process would continue to re-connect in a loop. That's the correct behavior as networking can go up/down.
This PR changes the original behavior. By adding `logErrorAndExit` in places where I see them could lead to cases when it's wasn't ideal to log & exit, eliminating existing re-load capabilities that happens on `ST2_COMMANDS_RELOAD_INTERVAL`.

scripts/stackstorm.js
```diff
@@ -409,22 +420,26 @@ module.exports = function(robot) {
...
-      source.close();
-    });
+
+    if (_stream) {
+      _stream.removeAllListeners();
+      _stream.close();
+    }
+
+    robot.shutdown();
+    process.exit(1);
```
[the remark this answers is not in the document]
blag: The original code didn't work properly. It created a new EventSource object (eg: a new websocket to StackStorm, I think), and then immediately shut it down and closed it.

Here we save the stream and shut it down if has been initialized.

Do you know why the module returns the `stop` function?

scripts/stackstorm.js
```diff
@@ -358,25 +368,26 @@ module.exports = function(robot) {
...
-        // TODO: squeeze a little bit more info out of evensource.js
-        robot.logger.error('Stream error:', err);
-      };
-      source.addEventListener('st2.announcement__chatops', function (e) {
+    robot.error(logErrorAndExit);
+
+    api_client.stream.listen().then(function (stream) {
+      _stream = stream;  // save stream for use in stop()
+      stream.on('error', function (err) {
+        logErrorAndExit(err);
```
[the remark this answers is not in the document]
blag: Fixed, as I think this is reasonable and the EventSource (stream) attempts to reconnect automatically.

scripts/stackstorm.js
blag: I've updated this to not exit immediately, and to simply emit an informative error about what's going on. The EventSource object should attempt to reconnect automatically.

scripts/stackstorm.js
```diff
@@ -358,25 +368,26 @@ module.exports = function(robot) {
...
-    } catch (e) {
-      robot.logger.error("Unable to decode JSON: " + e);
-      robot.logger.error(e.stack);
-      res.send('{"status": "failed", "msg": "An error occurred trying to post the message: ' + e + '"}');
+      res.send(JSON.stringify({
+        "status": "completed",
+        "msg": "Message posted successfully"
+      }));
+    } catch (err) {
+      logErrorAndExit(err, res)
```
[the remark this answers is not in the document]
blag: This is a bug in the original design, because `postDataHandler.postData(data);` sent data asynchronously (at least for Slack and MS Teams it does), so the catch block was simply never run. In the StackStorm web UI, it looked like the message was sent successfully, when it may not have been.

I have modified this to simply emit an error message letting the user know that the execution results may not be accurate. This is more in line with the previous code, but still probably isn't ideal.

scripts/stackstorm.js
```diff
@@ -236,8 +248,8 @@ module.exports = function(robot) {
         robot.logger.info(command_factory.st2_hubot_commands.length + ' commands are loaded');
       })
       .catch(function (err) {
-        var error_msg = 'Failed to retrieve commands from "%s": %s';
-        robot.logger.error(util.format(error_msg, env.ST2_API_URL, err.message));
+        robot.logger.error('Failed to retrieve commands from ' + env.ST2_API_URL);
+        logErrorAndExit(err);
```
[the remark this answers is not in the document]
blag: Reverted to previous behavior with a better error message. Although this behavior as well needs to be improved.

blag:
@armab Sorry, I wasn't clear on what you wanted earlier. I have fixed the code where you have pointed it out. Please re-review.

This code needs to handle reconnecting better, but that is outside the scope of this PR.

tests/test-st2-invalid-auth.js
blag: This is a fair criticism. It is difficult to work on code without refactoring it to be better, but I'll try to keep PRs smaller and more constrained in the future.

I do think including tests alongside the fix is appropriate though.

CHANGELOG.rst
blag: Fixed.

scripts/stackstorm.js
blag: As noted, #172 is not a proper fix for this issue, and parts of this PR have been reverted to more closely match previous behavior.

scripts/stackstorm.js
```diff
@@ -126,16 +124,30 @@ module.exports = function(robot) {
...
   } else if (env.ST2_AUTH_TOKEN) {
-    api.setToken({ token: env.ST2_AUTH_TOKEN });
+    api_client.setToken({ token: env.ST2_AUTH_TOKEN });
+  }
+
+  function logErrorAndExit(err, res) {
+    if (err) {
+      robot.logger.error(err.stack);
+    }
+    if (res) {
```
armab: I found that second param `res` is only used in `robot.error` Hubot handler for uncaught exceptions.

Does it make sense to move entire `logErrorAndExit` code into `robot.error` handler itself?
And from the code where we want to exit just `throw err` as it was before. Looks better in the context of https://stackoverflow.com/a/40321400/4533625 (case 3) discussion.
It will be caught by `robot.error` then.

By throwing when we want to end an app (which will then call `robot.error`) I think allows any other bubbled up tasks to actually finish in a more safe way. It'll also stick with the previous code logic primitives. 

@blag What do you think?

scripts/stackstorm.js
```diff
@@ -357,25 +366,29 @@ module.exports = function(robot) {
...
-    } catch (e) {
-      robot.logger.error("Unable to decode JSON: " + e);
-      robot.logger.error(e.stack);
-      res.send('{"status": "failed", "msg": "An error occurred trying to post the message: ' + e + '"}');
+      res.send(JSON.stringify({
+        "status": "completed",
+        "msg": "Message posted successfully"
+      }));
+    } catch (err) {
+      robot.logger.error('Encountered an error when attempting to post ' +
```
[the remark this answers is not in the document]
blag: Agreed.

scripts/stackstorm.js
```diff
@@ -162,26 +174,26 @@ module.exports = function(robot) {
...
+    return auth_client.authenticate(env.ST2_AUTH_USERNAME, env.ST2_AUTH_PASSWORD)
       .then(function (token) {
         robot.logger.info('Token received. Expiring ' + token.expiry);
-        api.setToken(token);
-        client.on('expiry', authenticate);
+        api_client.setToken(token);
+        auth_client.on('expiry', authenticate);
       })
       .catch(function (err) {
-        robot.logger.error('Failed to authenticate: ' + err.message);
```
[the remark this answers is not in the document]
blag: That's fine. Let's do this:

```js
robot.logger.error('Failed to authenticate:');
logErrorAndExit(err);
```

That way we get the best of both worlds.

scripts/stackstorm.js
blag: I considered this and ultimately decided against it. We use `logErrorAndExit` in one more place than just the Hubot error callback, and I suspect that we will expand the use of it in further commits.

scripts/stackstorm.js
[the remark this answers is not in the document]
blag: @jinpingh Please implement this.

scripts/stackstorm.js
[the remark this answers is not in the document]
blag: @jinpingh Please implement this too.

scripts/stackstorm.js
blag: @jinpingh What do you think?

jinpingh: Did run test case posted on https://github.com/StackStorm/st2chatops/issues/124#issuecomment-494075130. With @blag yesterday fix,  for invalid `ST2_API_URL` , `ST2_STREAM_URL` and `ST2_AUTH_TOKEN` , `st2chatops` logs out error message and keep waiting.   Is this the result we wanted?

jinpingh:
Did run test case posted on StackStorm/st2chatops#124 (comment). With @blag yesterday fix, for invalid ST2_API_URL , ST2_STREAM_URL and ST2_AUTH_TOKEN , st2chatops logs out error message and keep waiting. Is this the result we that wanted?

armab:
@jinpingh To understand exactly what do you mean, can you please paste the commands to reproduce? I'm also looking at the test cases in this PR which helped everyone to be on the same page.

Also, "keep waiting" turned out to have diffs. In reported https://github.com/StackStorm/st2chatops/issues/124 the original issue was hanging our process in unrecoverable and waiting state. If reload loop happens (every 2 mins), - it's a different state.

jinpingh:
> @jinpingh To understand exactly what do you mean, can you please paste the commands to reproduce? I'm also looking at the test cases in this PR which helped everyone to be on the same page.
> 
> Also, "keep waiting" turned out to have diffs. In reported [StackStorm/st2chatops#124](https://github.com/StackStorm/st2chatops/issues/124) the original issue was hanging our process in unrecoverable and waiting state. If reload loop happens (every 2 mins), - it's a different state.

scripts/stackstorm.js
armab: I'm actually talking here about reverting initial design where we previously `throw`'ed error as a way to stop hubot execution.
In our case `robot.error` handler will automatically catch those exceptions as unhandled and then call `stop`.

Here is the diff:


```diff
- robot.error(logErrorAndExit);

- function logErrorAndExit(err, res) {
+ robot.error(function(err, res) {
    if (err) {
      robot.logger.error(err.stack);
    }
    if (res) {
      res.send(JSON.stringify({
        "status": "failed",
        "msg": "An error occurred trying to post the message:\n" + err
      }));
    }

    stop({shutdown: true});
- };
+ });

- logErrorAndExit('Failed to authenticate: ' + err.message);
+ throw 'Failed to authenticate: ' + err.message;
```

By technical reasons I think that raising `robot.error` handler which then calls `stop()` via `throw` instead of `logErrorAndExit` function would align better with the existing shutdown mechanics.

Can add here cases when users want to re-use `hubot-stackstorm` module and so adding ability to catch errors and re-define `robot.error` on their side would make sense.

jinpingh:
> > @jinpingh To understand exactly what do you mean, can you please paste the commands to reproduce? I'm also looking at the test cases in this PR which helped everyone to be on the same page.

@armab  The way to run testcases is to modify  `/opt/stackstorm/chatops/st2chatops.env` file for ` StackStorm settings`, then run `/opt/stackstorm/chatops/bin/hubot` 

I think at least `st2chatops` should be exit for invalid ST2 Configurations.  If `ST2Stream` up and down by other reason, then  `st2chatops` should try the best effort to recovery not just quit.

scripts/stackstorm.js
```diff
@@ -235,8 +248,7 @@ module.exports = function(robot) {
         robot.logger.info(command_factory.st2_hubot_commands.length + ' commands are loaded');
       })
       .catch(function (err) {
-        var error_msg = 'Failed to retrieve commands from "%s": %s';
-        robot.logger.error(util.format(error_msg, env.ST2_API_URL, err.message));
+        robot.logger.error('Failed to retrieve commands from "%s":\n%s', env.ST2_API_URL, err);
```
armab: This specific code refactoring change or improvement leads to the following regression in error reporting:

Before:
```
ERROR Failed to retrieve commands from "http://donating-angelfish-st2api:9101/": Unauthorized
```

After:
```
ERROR Failed to retrieve commands from "http://donating-angelfish-st2api:9101/":
[object Object]
```

tests/test-st2-invalid-auth.js
```diff
@@ -0,0 +1,168 @@
...
+      stop({shutdown: true});
+
+      done();
+    }).catch(function (err) {
+      console.log(err);
+      done(err);
+    });
+  });
+
+  it("should raise exception with bad API key", function(done) {
```
armab: Is this test description still actual?

As @jinpingh found, bad API key doesn't lead to uncaught exception and end of bot's life.

scripts/stackstorm.js
```diff
@@ -248,7 +250,12 @@ module.exports = function(robot) {
         robot.logger.info(command_factory.st2_hubot_commands.length + ' commands are loaded');
       })
       .catch(function (err) {
-        robot.logger.error('Failed to retrieve commands from "%s":\n%s', env.ST2_API_URL, err);
+        var error_msg = 'Failed to retrieve commands from "%s": %s';
+        robot.logger.error(util.format(error_msg, env.ST2_API_URL, err.message));
+        // Exit in the first time when loadCommands is called with invalid ST2_API_URL, ST2_API_KEY or ST2_AUTH_TOKEN.
+        if (opts.exitOnFailure) {
+          logErrorAndExit(err);
```
armab: @jinpingh This change just brings back the logic that we agreed is not a solution as it will stop on _any_ error that potentially could be recovered. @blag just reverted it in previous commits.

For example, what if server returned temporary unavailable error or networking went down? Stopping hubot on all of those cases is not a solution. In this specific case, we need to stop only on Unauthorized. Trying to catch granular errors is a better solution in this situation.

Please see again:
https://stackstorm.slack.com/archives/G0HP0ML0M/p1559872202069500?thread_ts=1559839097.055600&cid=G0HP0ML0M
https://stackstorm.slack.com/archives/G0HP0ML0M/p1559873365070300?thread_ts=1559839097.055600&cid=G0HP0ML0M

Please ask any questions if you need to.

armab: Alright, here is the log:
```
ERROR Failed to retrieve commands from "http://donating-angelfish-st2api:9101/": Unauthorized
ERROR Stream error: Event { type: 'error', status: 401 }
```
Can we stop hubot only if we see `401` or `Unauthorized`?

scripts/stackstorm.js
```diff
@@ -180,7 +180,8 @@ module.exports = function(robot) {
         auth_client.on('expiry', authenticate);
       })
       .catch(function (err) {
-        err.message = 'Failed to authenticate: ' + err.message;
+        // Exit from invalid ST2_AUTH_USERNAME or ST2_AUTH_PASSWORD.
+        robot.logger.error('Failed credential authenticate.');
```
armab: "Failed credential authenticate" looks like not the best wording for error message.

Why `Failed to authenticate` message is not enough, just one line above?

jinpingh: I was trying to tell end user what kind of authentication is failed (Not api key or token).

scripts/stackstorm.js
jinpingh: This is only called once during`st2chatops` initialization. Did test after `st2chatops` is up running, then issue `service st2stream stop`,  after few minutes  issue `service st2stream start`, and `st2chatops` is back up on line without  issue.

jinpingh: I can check if it is ` status: 401` issue then exit, but we will miss to catch invalid `ST2_API_URL`

armab: Talking about unresolved API urls, - to avoid breaking anything else and going back and forward in this PR I'm OK to keep that for now in it's previous state because DNS lookup could be a temporary failure. It's not a blocking loop issue and not unrecoverable error like it was in https://github.com/StackStorm/st2chatops/issues/124 case.

armab: @jinpingh If we exit on all errors only on first run, `ST2_AUTH_TOKEN` could expire, while `ST2_API_KEY` could be revoked or changed meaning we won't be able to catch Unauthorized errors on next re-loads and st2chatops keeps running even when i got Unathorized response.
Unauthorized issues are important to fix as many users had problems with that before.

Please revert to previous commit and add requested `401/Unauthorized` granular check as well as double-verify/improve the tests cases if needed, because we'll close https://github.com/StackStorm/st2tests/pull/170 by incorporating tests in this PR.

Let's please do it and finish this PR.

scripts/stackstorm.js
armab: `Failed credential authenticate.` message is confusing and not user-friendly at all.

If you want to mention explicitly initial auth method used, the following would work better:
```diff
- err.message = 'Failed to authenticate: ' + err.message;
+ err.message = 'Failed to authenticate with st2 username and password: ' + err.message;
```

However this is not accurate, because saying that we're not using token is not true.
I's something that's requested  via st2 username and password pair and really utilized in this specific block. Just look a few lines before: "Token received'.

jinpingh: These code are only called with credential authentication, this is why I want to put more special message here to distinguish token and API key authentication. Will changed to `Failed to authenticate with st2 username and password: ' + err.message;`

scripts/stackstorm.js
```diff
@@ -252,10 +251,6 @@ module.exports = function(robot) {
       .catch(function (err) {
         var error_msg = 'Failed to retrieve commands from "%s": %s';
         robot.logger.error(util.format(error_msg, env.ST2_API_URL, err.message));
-        // Exit in the first time when loadCommands is called with invalid ST2_API_URL, ST2_API_KEY or ST2_AUTH_TOKEN.
-        if (opts.exitOnFailure) {
-          logErrorAndExit(err);
```
armab: Our previous conversation was around this line.
So please check for `Unauthorized` and include `logErrorAndExit` here

jinpingh: `Unauthorized` is appeared when called with loadCommand and it is only called `st2chatops` initialization. If we want capture authentication issue all the time, it will be moved  `start` function

scripts/stackstorm.js
```diff
@@ -237,6 +251,9 @@ module.exports = function(robot) {
       .catch(function (err) {
         var error_msg = 'Failed to retrieve commands from "%s": %s';
         robot.logger.error(util.format(error_msg, env.ST2_API_URL, err.message));
+        if (err.status === 401 && err.message.includes("Unauthorized")) {
```
armab: What if by refactoring reasons backend returned `401` HTTP code and `Authorization Required` as a message?
I think `||` would be more appropriate here.

scripts/stackstorm.js
```diff
@@ -126,16 +124,30 @@ module.exports = function(robot) {
...
-    api.setKey({ key: env.ST2_API_KEY });
+    api_client.setKey({ key: env.ST2_API_KEY });
   } else if (env.ST2_AUTH_TOKEN) {
-    api.setToken({ token: env.ST2_AUTH_TOKEN });
+    api_client.setToken({ token: env.ST2_AUTH_TOKEN });
+  }
+
+  function logErrorAndExit(err, res) {
+    if (err) {
+      robot.logger.error(err.stack);
```
armab: Did you notice how this line prints the following:
```
ERROR undefined
```
as hubot log on most of the errors?

@jinpingh let's please fix this.

jinpingh: Sure, will check err.stack before log it

scripts/stackstorm.js
```diff
@@ -237,6 +251,9 @@ module.exports = function(robot) {
       .catch(function (err) {
         var error_msg = 'Failed to retrieve commands from "%s": %s';
         robot.logger.error(util.format(error_msg, env.ST2_API_URL, err.message));
+        if (err.status === 401 || err.message.includes("Unauthorized")) {
```
armab: :+1:

tests/test-st2-invalid-auth.js
```diff
@@ -0,0 +1,167 @@
...
+      stop({shutdown: true});
+
+      done();
+    }).catch(function (err) {
+      console.log(err);
+      done(err);
+    });
+  });
+
+  it("should raise exception with bad auth token", function(done) {
```
armab: @jinpingh We really needs these 2 tests to be helpful.

From what I see current test promises it "should raise exception with bad auth token", however what it really does is checking for "Loading commands" message in the log, without verifying if exception was raised or bot is terminated.

Please make sure these tests actually test that hubot is called to exit on invalid ST2 token or key.

scripts/stackstorm.js
armab: Recent change still makes little sense as it just hides an issue.

`err.stack` is not available in most of the error conditions we've seen.
But we want error reporting to actually work and report.

Please provide an alternate solution to show some useful information as error log if `err.stack` is not defined.

tests/test-st2-invalid-auth.js
jinpingh: Modified test cases for token and API key yesterday. It ran on my two different systems successfully but failed on Travis. Removed changes and will work on that.

scripts/stackstorm.js
jinpingh: For current code, all log err messages and extra information are printed before  call `logErrorAndExit`, this is why did not put extra information here.

armab: So based on that, should developer always explicitly log the error before actually calling `logErrorAndExit`? Why error is not logged from the method itself if it's naming is `logError...`?

This puts a lot of knowledge burden on someone to who uses `logErrorAndExit` method and its corner cases and limitations. 
 
Additionally, remember that `logErrorAndExit` could be called from any unhandled exception because we're using it in `robot.error()`.

---

With that, I think we have enough reasons to actually _log_ error from the `logErrorAndExit` method.

tests/test-st2-invalid-auth.js
armab: Cool, thanks!

scripts/stackstorm.js
blag: > Add here cases when users want to re-use hubot-stackstorm module and so adding ability to catch errors and re-define robot.error on their side would make sense.

That isn't quite how the new code works.

Calling `robot.error` simply adds a callback function to the callback chain for the `error` event. See [this code](https://github.com/hubotio/hubot/blob/master/src/robot.js#L194) in Hubot for more.

If our users want to change or remove any of the error callbacks, they can still do so, they just have to manipulate `robot.errorHandlers` themselves. It's not as simple as assigning a function to `robot.onerror` anymore.

tests/test-st2-invalid-auth.js
```diff
@@ -152,7 +152,7 @@ describe("invalid st2 credential configuration", function() {
         }
       }
 
-      // Check that it was called at some point with 'Failed to retrieve commands from'
+    /*  // Check that it was called at some point with 'Failed to retrieve commands from'
```
jinpingh: @blag On travis, not sure why first two error messages are not printed out.
```
[Tue Jun 11 2019 21:54:57 GMT+0000 (Coordinated Universal Time)] ERROR StackStorm event stream error: Event { type: 'error', status: 401 }
[Tue Jun 11 2019 21:54:57 GMT+0000 (Coordinated Universal Time)] ERROR Implicitly attempting to reconnect to StackStorm event stream.
[Tue Jun 11 2019 21:54:57 GMT+0000 (Coordinated Universal Time)] ERROR Failed to retrieve commands from "https://localhost/api": Unauthorized - ApiKey with key_hash=cddbf16c36e281d74efd98b8fc9467d0b68e304ae4b0e0026f9b8ec5a95043fef7393dddd2edb712c4ed499f4752f7e233991b8b9053958748a1be051c30a1a5 not found.```

File: scripts/stackstorm.js (hunk 14 above)
Comment (blag) [reply scripts/stackstorm.js]: Making the hubot error handler an unnamed lambda function, and then simply re-`throw()`ing exceptions is fine.

**Comment by blag** [comment]
@armab Sorry to keep pulling you back to this, but I figured out how to test this all, and I think it's finally in a mergeable state!

**Comment by blag** [comment]
Coverage from current `master`:

```
------------------------|----------|----------|----------|----------|-------------------|
File                    |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------------|----------|----------|----------|----------|-------------------|
All files               |    82.73 |    79.59 |    83.76 |    82.24 |                   |
 lib                    |    96.88 |    96.68 |    97.75 |    96.75 |                   |
  command_factory.js    |      100 |      100 |      100 |      100 |                   |
  format_command.js     |      100 |      100 |      100 |      100 |                   |
  format_data.js        |      100 |      100 |      100 |      100 |                   |
  post_data.js          |      100 |      100 |      100 |      100 |                   |
  slack-messages.js     |    88.89 |    81.58 |    91.67 |    87.64 |... 34,35,37,38,43 |
  slack_monkey_patch.js |    28.57 |       80 |       50 |    28.57 |    17,18,19,28,29 |
  utils.js              |      100 |      100 |      100 |      100 |                   |
 scripts                |    41.81 |    37.11 |    39.29 |    41.81 |                   |
  stackstorm.js         |    41.81 |    37.11 |    39.29 |    41.81 |... 04,405,406,431 |
------------------------|----------|----------|----------|----------|-------------------|
```

Coverage from this branch:

```
------------------------|----------|----------|----------|----------|-------------------|
File                    |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------------|----------|----------|----------|----------|-------------------|
All files               |     87.9 |    86.34 |    86.67 |    87.56 |                   |
 lib                    |       99 |    99.57 |    98.86 |    98.96 |                   |
  command_factory.js    |      100 |      100 |      100 |      100 |                   |
  format_command.js     |      100 |      100 |      100 |      100 |                   |
  format_data.js        |      100 |      100 |      100 |      100 |                   |
  post_data.js          |      100 |      100 |      100 |      100 |                   |
  slack-messages.js     |      100 |      100 |      100 |      100 |                   |
  slack_monkey_patch.js |    28.57 |       80 |       50 |    28.57 |    17,18,19,40,41 |
  utils.js              |      100 |      100 |      100 |      100 |                   |
 scripts                |    59.28 |    58.56 |    53.13 |    59.28 |                   |
  stackstorm.js         |    59.28 |    58.56 |    53.13 |    59.28 |... 96,425,461,462 |
------------------------|----------|----------|----------|----------|-------------------|
```
👍

File: scripts/stackstorm.js (hunk 23 above)
Comment (blag) [reply scripts/stackstorm.js]: I removed `logErrorAndExit` entirely, since it didn't fit the semantics of what we wanted, so this conversation is obviated/resolved.

File: scripts/stackstorm.js (hunk 22 above)
Comment (blag) [reply scripts/stackstorm.js]: It doesn't make sense to handle an unauthorized error differently than another non-connection error, so I had it always handle it this way. The current code looks like (`function loadCommands`):

```js
      return (...).then(function () {
        ...
      })
      .catch(function (err) {
        var error_msg = 'Failed to retrieve commands from "%s":\n%s';
        err.stack = util.format(error_msg, env.ST2_API_URL, err.stack);
        robot.logger.error(err.stack);
        throw err;
```

This allows the `loadCommands` promise to log the error before it rejects, and then rejects are either ignored (`function start`):

```js
      // Add an interval which tries to re-load the commands
      commands_load_interval = setInterval(
        // Lambda function that swallows errors so they don't propagate up the stack
        // (errors are still logged within loadCommands itself)
        function () {
          loadCommands().catch(function (err) {});
        },
        (env.ST2_COMMANDS_RELOAD_INTERVAL * 1000));
```

or the rejection is simply passed up the stack (`function start`):
```js
      // Initial command loading
      return loadCommands();
```

and handled elsewhere (init code):

```
  // Authenticate with StackStorm backend and then call start.
  return authenticated.then(function () {
    return start().then(function () {
      return stop;
    });
  }).catch(function (err) {
    // Since start is only called once when the plugin is loaded, bail
    // loudly and harshly if loadCommands() isn't successful
    stop({shutdown: true});
    throw err;
  });
```

PR #169, opened 2019-06-04
Further tweaks for #168

blag:
Final tweaks for #168. This should be merged into the `issue-124/exit_on_failures` branch before #168 is merged in. That PR is looking pretty good, but there were a lot of places we can/should refactor, and some places we absolutely needed to refactor so we can test the new changes.

This PR:

* Adds large comment in `slack_monkey_patch.js` for future developers to help determine if the monkey patching is still necessary.
* Removes `Object.assign()` polyfill from `lib/slack-messages.js` - **this requires node.js version > 8.3.0**
* Breaks out the dummy `Logger` out from `tests/dummy-robot.js` into its own module
* Add `robot_name` argument to dummy `Robot` mock, to better mirror actual `Robot` constructor
* Rename `api` and `client` variables (both `st2client.js` instances) to `api_client` and `auth_client` in `scripts/stackstorm.js`
* Coalesce multiple `var` declarations into one
* Combine and rename `hubotErrorCallback` and `exitProcessWithLog` into `LogErrorAndExit`, which is now used to handle uncaught exceptions within Hubot, and is explicitly called when we cannot recover from re/connection attempts
* Refactor `loadCommands` to not accept an options object parameter, since we should always simply die if we can't connect or load commands
* Move `promise = authenticate();` into an `else` block to prevent it from running while the robot is stopping
* Use array unpacking (`var [var1, var2, var3] = [1, 2, 3];`) since we can now that we're using Node.js > 8
* Refactor (eventsource) stream error handling to properly register a callback instead of overwriting the `onerror` handler
* Remove ST2 stream error handler since we now have an `uncaughtException` handler for hubot
* Rename eventsource's `source` parameter to `stream`, since that makes slightly more sense
* Close down the eventsource stream instead of creating a new one that immediately gets shut down
* Add a few unit tests to test if hubot is stopped if certain environment variables aren't valid

### DIFF
--- a/tests/dummy-logger.js
+++ b/tests/dummy-logger.js
@@ -17,21 +17,44 @@ limitations under the License.
 
 "use strict";
 
-var Logger = require('./dummy-logger.js');
 
-function Robot(name, adapter, enable_logging) {
-  this.logger = new Logger(enable_logging);
-  this.name = name;
-  this.commands = [];
-  this.adapter = adapter;
 
-  this.messageRoom = function(recipient, data) {
-    return;
-  }
+function Logger(enabled) {
+  this.enabled = enabled;
+  this.logs = {
+    error: [],
+    warning: [],
+    info: [],
+    debug: []
+  };
 
-  this.emit = function(event, data) {
-    return;
-  }
+  this.error = function(msg) {
+    if (!this.enabled) {
+      return;
+    }
+    this.logs.error.push(msg);
+  };
+
+  this.warning = function(msg) {
+    if (!this.enabled) {
+      return;
+    }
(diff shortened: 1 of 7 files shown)


tests/dummy-logger.js
```diff
@@ -0,0 +1,60 @@
...
+  this.enabled = enabled;
+  this.logs = {
+    error: [],
+    warning: [],
+    info: [],
+    debug: []
+  };
+
+  this.error = function(msg) {
+    if (!this.enabled) {
```
jinpingh: Is following logical simpler?
``` 
if (this.enabled) {
   this.logs.error.push(msg);
}
```

Issue #158, opened 2018-08-17
Not all aliases reaching hubot-stackstorm


mbrannigan:
We have approximately 135 aliases:

```
# st2 action-alias list -j -a all | jq '.[].formats[]' | egrep "^\""  | wc -l
135
```

Not all of them are being made available to hubot. 

When dumping `Raw actions from ST2 API` during `loadCommands` in `/opt/stackstorm/chatops/node_modules/hubot-stackstorm/scripts/stackstorm.js`, it was determined that about 100 aliases were making it.

mbrannigan:
@blag thanks for helping out with this!  If you need any attachments / additional information, please let me know.

LindsayHill:
> about 100 aliases were making it

I have a strong suspicion that this is the same fundamental issue as https://github.com/StackStorm/st2/issues/4034

Changing https://github.com/StackStorm/st2/blob/master/conf/st2.conf.sample#L34 _might_ work.

mbrannigan:
Updated `st2.conf` to set `max_page_size` to `300` in the `api` section and restarted st2 with `st2ctl restart`.

https://gist.github.com/mbrannigan/1a0919e8d488402be6048092ccdb4d9a

`unpause` still isn't listed.

mbrannigan:
After cleaning up the gist a bit:

```
cat bar2 | jq -r '.[].formats' | egrep " \"" | grep ".*pause sidekiq"
  "pause sidekiq queues (on|in|for|with|using) environment {{ environment }}"
```

Validates that `unpause sidekiq` is still not listed.

```
cat bar2 | jq -r '.[].formats' | egrep " \"" | wc -l
148
```

148 aliases now listed during `loadCommands`

LindsayHill:
Validated, this is related to: 
https://github.com/StackStorm/st2/issues/4034

The "148" number is a red herring. There are 100 **aliases** returned, but those aliases have an average of > 1 format per aliases. So counting _formats_ gives us 148 unique formats, but there are really only 100 aliases.

Proper fix requires proper pagination handling with st2client.js

mbrannigan:
The following shows the number of aliases returned by the API

[...]

PR #176, opened 2019-06-12
Support pagination for action alias.

jinpingh:
Fixes https://github.com/StackStorm/hubot-stackstorm/issues/158 
Setup `limit=-1` to retrieve all the action alias records.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,10 @@ Changelog
 in development
 --------------
 
+0.9.4
+-----
+* Add pagination support for action aliases - fixes #158 (bug fix)
+
 0.9.3
 -----
 * Add initial support for Microsoft Teams via BotFramework (new feature)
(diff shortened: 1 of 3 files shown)


blag:
Please bump the required `st2client` version to `1.1.2` in [`package.json`](https://github.com/StackStorm/hubot-stackstorm/blob/master/package.json#L30) before merging this.

CHANGELOG.rst
```diff
@@ -5,6 +5,10 @@ in development
 --------------
 * Exit hubot on unhandled promise rejections and uncaught exceptions, usually caused by unrecoverable errors (bug fix)
```
armab: @jinpingh Because this change was already in `master`, changelog record should be moved under `0.9.4` during the merge with master/rebase process.

armab: Please pay more attention next time when syncing with `master`.

armab: cc @blag

blag: This one is on me - I noticed that and didn't change it, since I figured the issue that it is a part of is not quite "all the way" fixed. As far as I'm aware though, the changelog is only going to be read by humans, so the changelog being "wrong" isn't going to break any software.

I will be more strict about this in the future.

PR #180, opened 2019-06-21
Move babel-eslint to devDependencies instead of dependencies

blag:
Should help reduce the size of the diff of `npm-shrinkwrap.json` in https://github.com/StackStorm/st2chatops/pull/128.


blag:
Added a dashes line under the `0.9.6` version string to fix RST syntax from #183.

blag:
@armab Added an entry to the changelog. 👍

PR #181, opened 2019-06-27
(WIP) Remove undocumented ST2_AUTH_TOKEN from st2chatops

jinpingh:
`ST2_AUTH_TOKEN` is not documented in `st2chatops.env` and because the token either has a long expiry time or the expiry time is so short the ChatOps service is restarted all the time to pick up the config changes. It is no point to keep it, therefore remove this auth method from hubot-stackstorm entirely.

### DIFF
--- a/scripts/stackstorm.js
+++ b/scripts/stackstorm.js
@@ -52,9 +52,6 @@ env.ST2_WEBUI_URL = env.ST2_WEBUI_URL || null;
 env.ST2_AUTH_USERNAME = env.ST2_AUTH_USERNAME || null;
 env.ST2_AUTH_PASSWORD = env.ST2_AUTH_PASSWORD || null;
 
-// Optional authentication token
-env.ST2_AUTH_TOKEN = env.ST2_AUTH_TOKEN || null;
-
 // Optional API key
 env.ST2_API_KEY = env.ST2_API_KEY || null;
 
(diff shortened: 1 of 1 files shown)


armab:
I don't see any significant benefit in removing token, but not opposted with the change itself.

To make it happen, please check other StackStorm repos if we use any ST2_AUTH_TOKEN, for example in tests, CI, Docker images or other possible places.

Additionally, don't forget about changelog record and package version change.

jinpingh:
> To make it happen, please check other StackStorm repos if we use any ST2_AUTH_TOKEN with chatops, for example in tests, CI, Docker images or other possible places.

Thanks for point it out and you had good point. Already found some test codes depend on ST2_AUTH_TOKEN. 
I am studying coffeescript and read https://github.com/StackStorm/hubot-stackstorm/pull/133 how hubot-stackstorm is refracted.  It came ST2_AUTH_TOKEN discussion and don't want too many changes in one activity.  I will leave this PR alone for now.

armab:
I found this PR pretty minimal and easy to change, I'm probably missing some background discussion around https://github.com/StackStorm/hubot-stackstorm/pull/133.

As for other dependent repos that rely on token in chatops, - it would be nice to see those PRs first.
It looks like we have a freedom of removing all the uses of ST2_AUTH_TOKEN as a step before merging this particular change.

jinpingh:
> I found this PR pretty minimal and easy to change, I'm probably missing some background discussion around #133.
> 
> As for other dependent repos that rely on token in chatops, - it would be nice to see those PRs first.
> It looks like we have a freedom of removing all the uses of ST2_AUTH_TOKEN as a step before merging this particular change.
T

Good point again, thanks!

PR #190, opened 2019-08-02
Fix scripts loading

blag:
Hubot loads all modules in `src/`, which includes `stackstorm_api.js`.

This means that the `StackStorm` class gets initialized without a `new` keyword, which means that its `this` variable is `undefined`, so this line:

```js
  self.robot = robot;
```

causes an error later on `TypeError: Cannot read property 'robot' of undefined`.

Moving the `stackstorm_api.js` module into `src/lib/` avoids this situation.

The `stackstorm.js` module properly initializes the `StackStorm` class with the `new` keyword, so it stays.

Imports are updated to reflect the moved module.


blag:
Yep, I'll add a few lines to the changelog when I make a release.

armab:
Changelog should come with a change you're doing to be tightly coupled with PR.

PR #187, opened 2019-07-19
Modernize and refactor stackstorm.js

blag:
This PR breaks up `src/stackstorm.js` into `src/stackstorm_api.js` and `src/stackstorm.js` (which is now just a wrapper).

Additionally, the `stackstorm_api.js` module is refactored into a `StackStormApi` JS "class", and the closed over global functions in the original anonymous closure have been broken out into their own constituent public methods.

# Notes #

The `stop` function was reverted to the previous functionality to ensure that this refactoring doesn't break anything. That change will land in a future release.

~The `stop` function was changed slightly. Instead of [opening a _new_ connection to ST2's event stream and then immediately closing it](https://github.com/StackStorm/hubot-stackstorm/blob/df94be01676d332c78843bfd00ab43d5fdf87aa0/src/stackstorm.js#L421):~

```js
    api_client.stream.listen().then(function (st2stream) {
      st2stream.removeAllListeners();
      st2stream.close();
    });
```

~it now [saves the event stream handler](https://github.com/StackStorm/hubot-stackstorm/blob/refactor-stackstorm.js-again/src/stackstorm_api.js#L424) and uses that to [close and stop the event stream handler](https://github.com/StackStorm/hubot-stackstorm/blob/refactor-stackstorm.js-again/src/stackstorm_api.js#L479):~

```js
  if (self.st2stream) {
    self.st2stream.removeAllListeners();
    self.st2stream.close();
  }
```

Before, `authenticate()` was run [conditionally](https://github.com/StackStorm/hubot-stackstorm/blob/df94be01676d332c78843bfd00ab43d5fdf87aa0/src/stackstorm.js#L440):

```js
  if (env.ST2_API_KEY || env.ST2_AUTH_TOKEN || env.ST2_AUTH_USERNAME || env.ST2_AUTH_PASSWORD) {
    // If using username and password then all are required.
    if ((env.ST2_AUTH_USERNAME || env.ST2_AUTH_PASSWORD) &&
        !(env.ST2_AUTH_USERNAME && env.ST2_AUTH_PASSWORD && env.ST2_AUTH_URL)) {
      throw new Error('Env variables ST2_AUTH_USERNAME, ST2_AUTH_PASSWORD and ST2_AUTH_URL should only be used together.');
    }
    authenticated = authenticate();
  }
```

And now that check is [separated from the `authenticate()` call](https://github.com/StackStorm/hubot-stackstorm/blob/refactor-stackstorm.js-again/src/stackstorm_api.js#L213):
```js
[...]

PR #186, opened 2019-07-17
Modernize directory structure and refactor chat adapters

blag:
# Move `scripts/stackstorm.js` and `lib/` into `src/` #

I created the `src/` directory and moved the code into it. This is done to separate the code from the repository "metadata" files, like `package.json` and test files.

# Refactor chat provider adapters into their own modules #

I broke up `post-data.js` and `format-data.js` into modules for their constituent chat providers. All of the provider-specific code is gathered up into one module, and all other modules are agnostic to the chat provider.

Thanks to inheritance, this greatly simplifies all adapters, because now they all inherit from the `DefaultMessagingHandler` and only override the functions they need to. This is especially true for the Slack-like adapters (Mattermost, Cisco Spark, and Rocketchat), so much so that consolidating and simplifying them down further can be done in the future (once we have a test system for them).

Breaking the adapters up should now also make it more obvious to casual users how to fix things for their provider adpater, and also how to create a new provider adapter from scratch. Hopefully this translates into more "drive by fixes".

Finally, because all of the adapters are broken up and simplified, we can (if we so choose) rewrite individual ones in coffeescript, which might be a bit more readable than Javascript and is reverse-compatible with Javascript. Rewriting them to use ES6 (along with `class` and "proper" inheritance with the `extends` keywords) is also a possibility.

# Rename `tests/` to `test/` #

It's a small change, but I moved the tests from being in `tests/` to `test/`. This makes `mocha` find the tests automatically.
* https://github.com/hubotio/hubot-mock-adapter
* http://operatorerror.org/2013/11/create-npm-package-for-hubot-script/
* https://github.com/hubot-scripts/hubot-example/tree/d82ade38971b8a22a7eb37aaffc49ccaafd5d05c

# Update `CHANGELOG.rst` to use ReStructured Text syntax #

We have been surrounding words with single backticks to indicate fixed-width fonts, which is fine for Markdown files, but doesn't have the same effect for ReStructured Text files. Since the changelog is an RST file, I updated the syntax to be compliant with RST.

# Additional Notes #

I changed the comments on the copyright notice lines to single-line comments so that `gulp lint` passes properly. Previously it did not consider the notices when they were within multi-line comments.

# Coverage Differential #

Test coverage went from:

```
------------------------|----------|----------|----------|----------|-------------------|
File                    |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------------|----------|----------|----------|----------|-------------------|
[...]

PR #191, opened 2019-08-03
Redo adapter inheritance

blag:
This PR is on top of #190.

This PR slightly refactors the adapter inheritance for Slack-like adapters.

It adds in an intermediary `SlackLikeAdapter` for the Slack-like providers. This should make it easier to use `DefaultAdapter` methods in subclasses, like with what is done in `SparkAdapter.formatData` and `SlackLikeAdapter.normalizeCommand`.

This makes things a bit easier:

```diff
-ThisAdapter.super_.prototype.customMethod.call(...)
+DefaultAdapter.customMethod(...)
```

I also removed the `normalizeAddressee` from the MS Teams adapter, since it was an override of [the same function from `DefaultAdapter`](https://github.com/StackStorm/hubot-stackstorm/blob/3a14cdd8ed73fb7f2c3647df215e02035065eebd/src/lib/adapters/default.js#L82) (and the MS Teams adapter inherits from `DefaultAdapter`). The two functions were identical, however, so it makes no sense to override that function. And since the two functions were identical, I didn't need to change the tests at all for this fix.

~Once #190 is merged, I will either rebase and merge this with GitHub, or I will manually rebase this branch on that one, force push, and merge it.~ Done.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,8 +7,10 @@ in development
 * Move babel-eslint dependency to devDependencies (bug fix)
 * Small refactor and more tests (for ``scripts/stackstorm.js``) (improvement)
 * Refactor chat providers into their own modules (improvement)
-* Modernize directory structure to be more consistent with other hubot plugins (improvement)
-* Split out the functionality of ``src/stackstorm.js`` into ``stackstorm_api.js`` and refactor it be a JS old style class with a wrapper (improvement)
+* Modernize directory structure to be more consistent with other hubot plugins
+  (improvement)
+* Split out the functionality of ``src/stackstorm.js`` into ``stackstorm_api.js`` and refactor it
+  be a JS old style class with a wrapper (improvement)
 
 0.9.6
 -----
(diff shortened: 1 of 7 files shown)


src/lib/adapters/msteams.js
```diff
@@ -89,11 +89,4 @@ MSTeamsAdapter.prototype.normalizeCommand = function (command) {
   return command;
 }
 
-MSTeamsAdapter.prototype.normalizeAddressee = function(msg) {
```
armab: What happened to this `MSTeamsApater` code block, why it's been removed?

blag: Good question, I left this out of the PR description.

The MS Teams adapter [inherits from `DefaultAdapter`](https://github.com/StackStorm/hubot-stackstorm/blob/3a14cdd8ed73fb7f2c3647df215e02035065eebd/src/lib/adapters/msteams.js#L28), and this code for `normalizeAddressee` in the MS Teams adapter is the exact same as the [`normalizeAddressee` function from `DefaultAdapter`](https://github.com/StackStorm/hubot-stackstorm/blob/3a14cdd8ed73fb7f2c3647df215e02035065eebd/src/lib/adapters/default.js#L82). Overriding the function was an oversight in #186.

I've updated the PR description to include this. I figured the PR tagline was sufficiently ambiguous to include this change as well.

blag:
I fixed a few things in the changelog, and added PR references to the most recent changes, including #190.

CHANGELOG.rst
```diff
@@ -4,24 +4,29 @@ Changelog
...
-* Small refactor and more tests (for ``scripts/stackstorm.js``) (improvement)
-* Refactor chat providers into their own modules (improvement)
-* Modernize directory structure to be more consistent with other hubot plugins (improvement)
-* Split out the functionality of ``src/stackstorm.js`` into ``stackstorm_api.js`` and refactor it be a JS old style class with a wrapper (improvement)
+* Move babel-eslint dependency to devDependencies [PR #180] (bug fix)
+* Small refactor and more tests (for ``scripts/stackstorm.js``) [PR #185] (improvement)
+* Refactor chat providers into their own modules [PR #186] (improvement)
+* Modernize directory structure to be more consistent with other hubot plugins [PR #186, PR #191]
+  (improvement)
+* Split out the functionality of ``src/stackstorm.js`` into ``stackstorm_api.js`` and refactor it
```
armab: Wasn't #190 a fix?

blag: My thinking was: #187 was an improvement (refactoring), #190 was a fix introduced in that refactoring.

So between the last release and the next release the combined result of #187 and #190 amounts to just an improvement, not an improvement+bugfix.

armab: Ideally to consider every PR as isolated change that should have dedicated Changelog record, instead of assuming that one PR is an addition to another.

PR #192, opened 2019-08-03
Fix Mattermost adapter

blag:
This is the fix for the Mattermost adapter from [this forum post](https://forum.stackstorm.com/t/st2chatops-with-mattermost-not-replying/711/25).

I have tested this manually with a minimal Mattermost server in a VM.

The [`slack-attachment` event](https://github.com/loafoe/hubot-matteruser/blob/aa2814a6594183535fddff8503bde35576bbd4a5/src/matteruser.coffee#L84) was removed from the hubot-matteruser dependency in [PR 71](https://github.com/loafoe/hubot-matteruser/pull/71), but our Mattermost adapter did not get updated to avoid using that.

I switched over to using the [`send` method from hubot-matteruser](https://github.com/loafoe/hubot-matteruser/blob/master/src/matteruser.js#L205).

Note: That methods accepts two parameters: `envelope` and a variadic parameter `strings`, but contrary to the name, the `strings` parameter also works just fine if each "string" is a message-like object, because it calls [`postMessage`](https://github.com/loafoe/mattermost-client/blob/master/src/client.js#L547) from mattermost-client, and that handles objects:

```js
    postMessage(msg, channelID) {
        const postData = {
            message: msg,
            file_ids: [],
            create_at: 0,
            user_id: this.self.id,
            channel_id: channelID
        };

        if (typeof msg === 'string') {
          postData.message = msg;
        } else {
          postData.message = msg.message;
          if (msg.props) {
            postData.props = msg.props;
          }
          if (msg.file_ids) {
            postData.file_ids = msg.file_ids;
          }
        }
```
<sup>(reformatted that code slightly to improve readability)</sup>

Once #190 and #191 are merged, I will either rebase and merge this with GitHub, or I will manually rebase this branch on #191, force push, and merge it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ in development
   (improvement)
 * Split out the functionality of ``src/stackstorm.js`` into ``stackstorm_api.js`` and refactor it
   to be a JS old style class with a wrapper [PR #187, PR #190] (improvement)
+* Fix the Mattermost adapter to use the ``adapter.post`` function instead of emitting the
+  ``slack-attachment`` event [PR #192] (bug fix)
 
 0.9.6
 -----
--- a/src/lib/adapters/mattermost.js
+++ b/src/lib/adapters/mattermost.js
@@ -17,6 +17,7 @@
 var env = process.env;
 var util = require('util');
 var utils = require('./../utils');
+var messages = require('./../slack-messages');
 var SlackLikeAdapter = require('./slack-like');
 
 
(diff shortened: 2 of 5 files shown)


blag:
@armab I added a change to the changelog. Since I won't ever be merging the changes from #193 directly into `master`, the changelogs will differ, but this and #193 both fix the same thing. Let me know if this is acceptable to you.

armab:
As mentioned in #193, `master` should have in its Changelog record about every release that happened in the project. It's not OK to compose `master` Changelog like `0.9.7` release never happened.

So let's please mention in Changelog a few notes about 0.9.7 for our users to have this information visible in history. This is not any worse then split-brain with patch and I hope its first and last time we're doing this.

Additionally, are you going to release `v0.10.0`? Please prepare the package version and changelog for that as well in PR.

blag:
I closed #193. We won't do a split brain release.

I have a version bump local branch ready to go, but I'd like to merge this and then rebase that branch on top of `mater` before I push it and open a PR. This is so I don't have to do a rebase and force push (since I know you don't like those).

armab:
Sure, works for me.

PR #193, opened 2019-08-05
Fixes for v0.9.7

blag:
Pulling the fixes from #192 into the 0.9.7 branch.

This was manually tested separately from #192 with a minimal Mattermost installed on VM.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,11 +5,12 @@ in development
 --------------
 
 0.9.6
+-----
 * Don't consider failed alias execution as a critical reason to exit hubot (bug fix)
 
 0.9.5
 -----
-* Exit hubot on invalid, expired `ST2_API_KEY` / `ST2_AUTH_TOKEN` or any other Unauthorized response from st2 server (bug fix)
+* Exit hubot on invalid, expired ``ST2_API_KEY`` / ``ST2_AUTH_TOKEN`` or any other Unauthorized response from st2 server (bug fix)
 * When st2 username/password is used, re-generate st2 auth token in advance, giving enough time for request to complete, st2client.js (bug fix)
 
 0.9.4
@@ -24,7 +25,7 @@ in development
 
 0.9.2
 -----
-* Rename ENV variable `ST2_API` -> `ST2_API_URL` for consistency, keep `ST2_API` for backwards compatibility (improvement)
+* Rename ENV variable ``ST2_API`` -> ``ST2_API_URL`` for consistency, keep ``ST2_API`` for backwards compatibility (improvement)
 
 0.9.1
 -----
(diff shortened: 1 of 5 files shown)


armab:
Don't forget to apply respective Github labels for better Issues organization.

armab:
Also after merging this, how do you plan to sync-up changes back to the current master?

blag:
> I get the idea of releasing an isolated fix to the user, but let's please avoid detached/split releases like that in future.
>
> Instead we should try to release more frequently from master.

I was following [SemVer](https://semver.org/#summary):

> PATCH version when you make backwards-compatible bug fixes.

AIUI, this is the same approach we take in st2, where we have feature and bugfix releases.

> Instead we should try to release more frequently from `master`.

I don't quite understand how that would help this situation, can you explain a bit more?

> Also after merging this, how do you plan to sync-up changes back to the current master?

I don't. The `v0.9.7` branch will only receive bugfixes until we release a version of st2chatops that contains hubot-stackstorm version 0.10.0 or higher.

PR #192 is the same fix for `master`. Please give #192 and #193 a review when you have a chance. Once that's merged I'll have a version bump PR for version 0.10.0 and release that as well.

armab:
It may look similar, however both git history and results are different.

For the st2 repo we maintain the versioned "stable" branches, which is not the case for repositories like `hubot-stackstorm`. In `st2` every change is first merged into `master` and then cherry-picked into respective branch, but not like we do here with git brain split situation when `v0.9.7` is not a subset of master.

After all you'll still need to include in master Changelog for the `0.9.7` release, and master history shows different git log and order for the fixes.
To avoid this, ideally to release more frequently from `master` once the logically sufficient change was made. And so we'd release `v0.10.0` with refactoring, `v0.10.1` with some other fixes and `v0.10.2` with Mattermost bugfix, all based on master, preserving consistency.

blag:
Instead of having a bugfix release branch around indefinitely, I think users will need to manually apply these two commit diffs:

https://github.com/StackStorm/hubot-stackstorm/pull/193/commits/1cb50ac344a78c65e42f7a0ff88843a8952f8645
https://github.com/StackStorm/hubot-stackstorm/pull/193/commits/d89bd18c1048a36f563b0c9fc8d73b54c52ec0f2

Closing.

armab:
For the temporary fix before we officially release v3.2 for st2 and st2chatops, installing hubot-stackstorm `v0.10.0` would be fine too.

blag:
Just a note: hubot-stackstorm `v0.10.0` does contain a giant amount of refactoring. The intent with this branch was to get users of older versions of hubot-stackstorm (eg: `v0.9.6` and below) a fix/patch that did not include the refactoring work going on.
¯\\\_(ツ)\_/¯

armab:
That's hopefully OK, considering that refactoring is non-breaking and didn't really introduce any changes.
After all we're going to rely on that in next st2 release.

PR #200, opened 2020-02-07
Fix issue with webex adapter

viveksyngh:
Fix issue with webex adapter

Closes https://github.com/StackStorm/hubot-stackstorm/issues/198

Also, created this PR with `hubot-spark` adapter.

https://github.com/tonybaloney/hubot-spark/pull/23

PR #202, opened 2020-04-11
Version 0.10.2

blag:
Bump version to `0.10.2` and update changelog.

* [x] This should be merged after #201.

~I also properly shrinkwrapped the package to `npm-shrinkwrap.json`, so that's a _lot_ smaller and a lot more reasonable now.~

I completely removed `npm-shrinkwrap.json`, because it doesn't allow Travis CI to install `devDependencies` that are necessary to run tests. Then I did some searching around and found [this blog](https://medium.com/@hossam.hilal0/package-json-vs-package-lock-json-vs-npm-shrinkwrap-json-33fcddc1521a) that walks through `package.json`, `package-lock.json`, and `npm-shrinkwrap.json`. It's not entirely clear from that blog whether or not we should be using `package-lock.json` or `npm-shrinkwrap.json`, because he doesn't explicitly cover what would be appropriate when publishing a plugin type of package on NPM.

However, I googled "hubot plugin npm-shrinkwrap.json" and the only search result was...this very repository. Starting to suspect we had been doing the wrong thing, I searched for "hubot plugin package-lock.json" and got a lot more search results:
* [Hubot-BotFramework](https://github.com/microsoft/BotFramework-Hubot/blob/master/package-lock.json)
* [hubot-rocketchat](https://github.com/RocketChat/hubot-rocketchat/blob/master/package-lock.json)
* [hubot-youtube](https://github.com/hubot-scripts/hubot-youtube/blob/master/package-lock.json)
* [hubot-home-assistant](https://github.com/home-assistant/hubot-home-assistant/blob/master/package-lock.json)

And most tellingly, none of those plugins had `npm-shrinkwrap.json` files!

Due to that, this PR:

* [x] removes our `npm-shrinkwrap.json` file
* [x] tweaks `.gitignore`
* [x] regenerated the `package-lock.json` file

I also figured out that the `package-lock.json` file saves `dev: true` for all `devDependencies`, so we should be able to ignore those dependencies when updating this in st2chatops, and also when generating that project's package lock file. If anybody knows any JS and/or NPM gurus that can double check this decision, please contact them.

I also double checked that our CI builds all actually did pass all of the tests for the node versions that we care about, since they "passed" but should have failed in #200. There's a failure with Node v12, but we don't support that yet so it's not a big deal. We do test it though, and include it as an allowed failure.


blag:
Published to NPM. :+1:

PR #203, opened 2020-04-21
Update st2client.js to 1.2.2

blag:
This PR updates st2client.js from `1.1.3` to version `1.2.2`, so it pulls in the changes from PR StackStorm/st2client.js#76.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 in development
 --------------
 
+0.10.3
+------
+
+* Update st2client.js to 1.2.2 [PR #203] (change)
+
 0.10.2
 ------
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -371,9 +371,12 @@
       "dev": true
     },
     "axios": {
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-0.7.0.tgz",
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -4,6 +4,11 @@ Changelog
 in development
 --------------
 
+0.10.3
+------
+
+* Update st2client.js to 1.2.2 [PR #203] (change)
```
armab, line 10: Ideally to indicate in the changelog that version included an update to potentially vulnerable dependencies

blag: I generally try to avoid mentioning this, since people may scan changelogs for security updates, figure out how to attack older ST2 instances, and then start attacking older ST2 installations. The Linux kernel ran into this issue awhile back IIRC.

armab: I disagree.

Users _should_ know whether specific versions of the released software had any potential security issues, fixes, mitigations or not. This could be a driving factor for them to update or not.

I don't think it's a common practice to hide this kind of information behind the vague changelog. 

---

For example here is pip changelog https://pip.pypa.io/en/stable/news/#id302:
> Upgrade the bundled copy of requests to 2.6.0, fixing CVE-2015-2296.

This changelog is related to updating pip's dependency `requests` which had a security issue (https://nvd.nist.gov/vuln/detail/CVE-2015-2296).

PR #204, opened 2020-06-05
Bump version to 0.10.4 (update st2client to 1.2.3)

blag:
* Update `st2client` to 1.2.3
* Install `mversion` as a developer dependency so we can use it to bump versions in the future (like what the README says to do)
* Updated the changelog
* Bumped version to 0.10.4 with `mversion`


armab:
cc @bgaeddert

PR #214, opened 2020-12-30
Handle slack injected nbsp

cognifloyd:
copy/paste in slack proves problematic
because sometimes slack replaces a space
with the Latin1 non breaking space char /xA0

this cleans that up so that alias commands
and extra args can be appropriately recognized

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,7 +9,7 @@ in development
 ------
 
 * Clean slack-injected latin1 nbsp character (/xA0) from command to improve command recognition.
-  Slack will sometimes replace a space with /xA0, ofen near something that renders as a link.
+  Slack will sometimes replace a space with /xA0, often near something that renders as a link.
   Now, when someone copies that command, they can paste it and still have hubot recognize it.
   (improvement) #214
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,14 @@ in development
 --------------
 * Update ``lodash`` and ``st2client`` dependencies (improvement)
 
+0.12.0
+------
+
+* Clean slack-injected latin1 nbsp character (/xA0) from command to improve command recognition.
+  Slack will sometimes replace a space with /xA0, ofen near something that renders as a link.
+  Now, when someone copies that command, they can paste it and still have hubot recognize it.
+  (improvement) #214
+
 0.11.2
 ------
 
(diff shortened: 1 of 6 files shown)


blag:
Can you add tests for this?

cognifloyd:
Added tests and rebased on master.

src/lib/adapters/slack-like.js
```diff
@@ -56,6 +56,11 @@ SlackLikeAdapter.prototype.normalizeCommand = function(command) {
   command = command.replace(/\u2018/g, '\u0027');
   // replace right single quote with regular apostrophe
   command = command.replace(/\u2019/g, '\u0027');
+  // replace non-breaking space with regular space
+  // first the unicode variant, so we don't cut it in half,
+  // then the latin1 variant (thanks to slack)
+  command = command.replace(/\u00A0/g, ' ');
+  command = command.replace(/\xA0/g, ' ');
```
blag, line 63: NAK.

The `SlackLikeAdapter` is used by any adapter that uses a Slack-like interface, but not all of them may have this issue. Since you are only adding tests for the Slack adapter and not all of the SlackLikeAdapters, this would probably only apply to the Slack adapter itself. That means that the better place for this would be in `src/lib/adapters/slack.js`.

Within the SlackAdapter, I would do it this way:

```js
SlackAdapter.prototype.normalizeCommand = function (command) {
  SlackLikeAdapter.prototype.normalizeCommand.call(this, command);
  // replace non-breaking space with regular space
  // first the unicode variant, so we don't cut it in half,
  // then the latin1 variant (thanks to slack)
  command = command.replace(/\u00A0/g, ' ');
  command = command.replace(/\xA0/g, ' ');
  return command;
}
```

I am writing this code off the cuff, so definitely test it out.

blag (changes requested):

src/lib/adapters/slack-like.js
cognifloyd: done. Plus added version bump and changelog entry

CHANGELOG.rst
```diff
@@ -5,6 +5,14 @@ in development
 --------------
 * Update ``lodash`` and ``st2client`` dependencies (improvement)
 
+0.12.0
+------
+
+* Clean slack-injected latin1 nbsp character (/xA0) from command to improve command recognition.
+  Slack will sometimes replace a space with /xA0, ofen near something that renders as a link.
```
blag, line 12:
```suggestion
  Slack will sometimes replace a space with /xA0, often near something that renders as a link.
```

blag (changes requested):
One last typo to fix in the changelog. Otherwise looks good, thanks!

cognifloyd:
LOL oops. Fixed

blag (approved):

PR #218, opened 2021-05-11
Update node checks

amanda11:
Update to check against node  > 12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,6 @@
 language: node_js
 
 node_js:
-  - "6"
-  - "7"
-  - "8"
-  - "9"
   - "10"
   - "11"
   - "12"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -19,4 +21,6 @@ matrix:
     - node_js: "7"
     - node_js: "11"
     - node_js: "12"
+    - node_js: "13"
+    - node_js: "14"
```
blag, line 25:
```suggestion
```

We should force these to pass from now on.

blag (changes requested):
Looks good, although I'd like to minimize our test matrix as well (eg: remove tests for Node < 10) and force us to pass on more up-to-date pinned Node versions.

amanda11:
@blag - thanks for review. Shorted test matrix, am I ok to leave hubot-stackstorm version as before as not changed anything in the actual package?

blag:
For this change, that's correct - we do not need to bump and push a new version to NPM, as this is just testing changes.

If we update dependencies, like all of @dependabot's PRs, we should bump the version. But we can "batch" those together with respect to versioning - in other words, we can merge in multiple dependency updates before we bump the version and push it to NPM.

blag (approved):

PR #232, opened 2023-12-16
Upgrade node 20

rebrowning:
Upgrade to node 20, had to rewrite the gulp file to allow myself to run the tests with the newer version. Upgraded other libraries that did not cause unittests to break.

Ran the st2chatops repo with this package and am successfully able to use a slackbot, so I will consider these changes working (going to also open a PR


nzlosh:
As a side note: `err-stackstorm` is a Python chatops implementation.  If you're interested, you can read more about it here https://err-stackstorm.readthedocs.io/en/latest/

rebrowning:
> As a side note: `err-stackstorm` is a Python chatops implementation. If you're interested, you can read more about it here https://err-stackstorm.readthedocs.io/en/latest/

unfortunately this is about upgrading the package itself so that a more modern version of nodejs can be used (bringing it in line with node 20). Switching to a different chatops implementation is not an option.

nzlosh:
@rebrowning Tests have been switched from TravisCI to Github Actions which targets nodejs 14 to 20.  The tests are passing on master, but fail on your branch.

rebrowning:
@nzlosh I'm going to open another PR #234 to see if we can do without most of these changes and upgrades. I did this PR first but then st2chatops essentially just needed coffeescript pinned, so I'll see if that works here as well.

PR #233, opened 2024-01-05
Use github actions to test nodejs.

nzlosh:


.github/workflows/nodejs.yaml
```diff
@@ -0,0 +1,22 @@
+name: Node.js CI
+
+on: [push]
```
mamercad, line 3: How about:

```
on:
  push:
    branches: [master]
  pull_request:
  workflow_dispatch:
```

mamercad:
Perhaps remove [this](https://github.com/StackStorm/hubot-stackstorm/blob/master/.travis.yml) as well?

mamercad:
Needs an entry [here](https://github.com/StackStorm/hubot-stackstorm/blob/master/CHANGELOG.rst), right?

mamercad (approved):
🚢

PR #234, opened 2024-01-12
simplify the upgrade so we're backwards compatible

rebrowning:
see if we can just do this minimal change like we were able to do in st2chatops


nzlosh (approved):

nzlosh:
I've merged the MR, it looks good to me, however I just wonder why you need the other coffeescript pinning when tests were passing for nodejs 20 on master?

rebrowning:
@nzlosh  at this point I'd say consistency and to prevent any future upgrades breaking this unexpectedly (unless we intentionally upgrade coffeescript as well). st2chatops required the pin for coffeescript because of the name change that occurred (coffee-script -> coffeescript). It causes a lot of unexpected failures if an upgrade is attempted because any libraries that changed to the `coffeescript` name as a dependency are no longer pinned.

PR #238, opened 2025-03-12
Upgrade to Hubot 11

skiedude:
*These changes only support using Slack*

Due to multiple deprecations and archival of related chat backend providers, everything has been removed but Slack. 

1. Upgrade hubot dependency to 11
2. Upgrade nodejs dependency to 20, and only test against nodejs20
3. Removes all providers but slack
4. Removes old coffee script and other dependencies (required by previous version of hubot-slack)
5. Adds a threading feature I had provided in a separate PR https://github.com/StackStorm/hubot-stackstorm/pull/237
6. remove tests for all providers but slack
7. Update tests to import hubot src as async as its now ESM and this is still commonjs

These changes have been tested and running locally in our environments for 3 weeks now without issue.


nzlosh (approved):

package.json
```diff
@@ -22,17 +23,14 @@
   },
   "dependencies": {
     "cli-table": "<=1.0.0",
-    "coffee-register": "1.0.0",
-    "coffee-script": "1.12.7",
-    "coffeescript": "1.12.7",
```
cognifloyd, line 27: At work, we inject a coffeescript that adds rudimentary rbac for chatops. So, removing coffeescript is sad for me as it means extra work to replace that script (maybe compile to js?). But dropping it makes sense here since this repo doesn't make use of it.

cognifloyd (approved):

package.json
skiedude: Good to know, I just dropped it because it seemed to be a requirement of the old version of `hubot-slack`, now the new version doesn't require it so I removed it.

